PR #20, opened 2023-07-17
Add code-signing for Homebrew proposal

di:
This adds a proposal by myself and @woodruffw to add code-signing to the Homebrew ecosystem.

Edit: The goal here is to have a proposal from 'within' this WG that could be submitted to any potential funder. We plan to submit this to Alpha/Omega for funding first.


proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+
+For third-party taps hosted on GitHub, the signing flow would be identical: any bottle-building workflow (including reusing the official one) would sign for bottles.
+
+## Client-side signature verification in Homebrew
+
+The Sigstore signatures described above are submitted to each tap's object storage so that they can be verified by clients during the normal `brew install` flow.
+
+To verify a Sigstore signature, a particular `brew` client needs the following:
+
+- A trust proposition, i.e. a reason to trust the identity producing the signature. For Homebrew, this trust proposition comes from the tap's own identity: the act of performing `brew tap foo/bar` establishes that Sigstore identities tied to `github.com/foo/bar` can be trusted to sign for bottles in that tap. This applies to implicit `brew tap` invocations as well, e.g. `brew install foo/bar/baz` asserts that the user expects a signature on `baz` from the `foo/bar` GitHub identity.
```
haydentherapper, line 47: This is great! Love that identity can be automatically derived.

To confirm, there would be no support for user identities, only GitHub Actions identities?

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+- The local `brew` client looks for the "flag day" metadata and, if present and before the local system time, enforces signature verification for all bottles referenced by the tap.
+
+And at `brew install` time:
+
+- For each bottle resolved during a `brew install` invocation, the bottle must have an accompanying signature. If the accompanying signature is not present, then `brew install` must fail.
+- For each bottle and signature, `brew install` must internally reconstruct the _attestation_ described in the signing process above. This is then used as the cleartext input for signature verification. This input is then passed into the underlying Sigstore client along with its associated signature; the Sigstore client is responsible for performing both the signature verification and its associated identity verification.
+- If the underlying signature verification fails, then `brew install` must fail.
+
+Like with previous Homebrew features, an experimental feature-flagged variant of signature verification may be merged and enabled before it is enabled by default.
```
haydentherapper, line 63: For enabled by default, do we mean that if the flag day is set, then the client will always try to verify the signature? Or is the goal to require that all core bottles are signed?

jalseth:
This is great! One thing I would like to see added is a list of goals and non-goals. For example: This prevents attackers who can forge TLS certs for GitHub from tampering with artifacts, but does not prevent an upstream repo from being compromised and Homebrew CI from automatically building/signing/distributing a malicious package. The latter is partially addressed in this proposal, but would not be 100% addressed until a requirement that all Homebrew upstream sources are signed is added to the CI builder's logic. A future proposal may add a config option to enable a TAP owner to mark their TAP as 100% signed so all packages within the TAP must have signatures before Homebrew CI should build/distribute.

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Full source and build provenance with Homebrew and Sigstore
+
+Bottle signing and subsequent verification unlocks a potential subsequent improvement: because Homebrew performs source builds to produce bottles, Homebrew could also verify signatures for those source artifacts. That signature could then be _countersigned_ for by Homebrew's own bottle signatures, effectively producing a strong cryptographic association from the original source artifact (produced by a third-party upstream) to the Homebrew-produced binary artifact.
```
haydentherapper, line 67: Could this use the GitHub Actions generator for SLSA provenance (for build provenance, source provenance is a WIP track for SLSA)? - https://github.com/slsa-framework/slsa-github-generator

proposals/code-signing-for-homebrew.md
woodruffw: > Or is the goal to require that all core bottles are signed?

Yep, that's what I was thinking: my phrasing is a little messy here, but what I had in mind was "verification is optional while we incrementally update/rebuild all bottles with signatures, and once every core bottled is signed then everything *must* be signed after the flag day."

proposals/code-signing-for-homebrew.md
woodruffw: Yes! I think that's probably exactly what we want to do 🙂 -- I think the only outstanding big question there is around discoverability (i.e., how should Homebrew go from "here's my source archive" to "I know that there's a SLSA provenance available for this source"?)

proposals/code-signing-for-homebrew.md
woodruffw: Yep, that was my thought for the MVP: the Ruby DSL specified below in the proposal would give us the flexibility to support non-GHA identities in the future, but the 99.9% case (IMO) doesn't need them.

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+
+The high-level flow for bottle signing:
+
+1. Existing bottle-building workflows will gain a signing step, which will only run as a finalization step for bottles that appear on `master` (i.e., not just the temporary bottles that are built as part of the normal PR lifecycle).
+2. The signing step is responsible for producing a Sigstore signature that _attests_ to the bottle. This should include, at minimum, the bottle's SHA256 digest and the bottle's filename. The digest is necessary to produce a proof that the signature is for a particular bottle's contents, and the filename is necessary to produce a proof that the signature is intended to accompany a particular install step (i.e. `brew install foo` should not verify if an attacker substitutes `foo`'s bottle for a correctly signed bottle of bar, or even a different version of foo than the resolver produces).
```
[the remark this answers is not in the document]
woodruffw: I think @di will need to update the PR here, but to clarify: the thing being signed over is an attestation over (at least) the following:

1. A digest of the bytes for the bottle archive, i.e. `H(read("some-bottle-xxx.tar.gz"))`
2. The bottle's base filename (to prevent domain confusion between signatures/versions of bottles)

My thought was to define that as a custom payload, since Homebrew's bottles aren't *actually* OCI manifests -- they just happen to be stored that way in `homebrew-core`'s storage, since they reuse GitHub Packages.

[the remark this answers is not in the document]
di: (@woodruffw I can update but won't be able to get to this today, if you want to make a PR against this branch I'll be happy to merge changes that way)

proposals/code-signing-for-homebrew.md
di: Maybe @steiza can chime in on this specific topic?

proposals/code-signing-for-homebrew.md
haydentherapper: Yea, it'd be worth a brief mention that the focus is on CI identities.

proposals/code-signing-for-homebrew.md
steiza: When we built the attestation store for npm, we had in mind that we could extend it so GitHub repos could also store and fetch attestation documents. We're thinking of exposing a public API where you supply the repo and artifact hash, and get back the associated attestation document.

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: https://github.com/ossf/wg-securing-software-repos/pull/21/commits/fc958da7a7ce99330ec7a7ed711d0379b89ccb89

proposals/code-signing-for-homebrew.md
woodruffw: https://github.com/ossf/wg-securing-software-repos/pull/21/commits/0f9675a427973d84ee56f84c821b12dd526c987c

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
lukehinds: Could we not use cosign style blob signing like they do for wasm and other blobby things?

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+    oidc_issuer "<...>"
+  end
+end
+```
+
+For source artifacts hosted on GitHub and provided (along with their signatures) through GitHub releases, the first option will be sufficient: Homebrew can re-derive the expected signing identity from the `url`'s `user/repo` component.
+
+For source artifacts hosted on other services (or that are signed with a non-GitHub-Actions identity), the second option allows the formula to explicitly specify the signature's URL and associated identity.
+
+For formulae with verifiable source artifacts, the attestation described above can be extended to include the verified signature, resulting in a countersignature. This countersignature can be verified by `brew` (through a Sigstore client) using the same basic client-side signature verification process described in the task above, making this a progressive enhancement to "bottle-only" signatures.
```
joshuagl, line 101:
```suggestion
For formulae with verifiable source artifacts, the signed bottle attestation described above can be extended to include the verified source signature, resulting in a countersignature. This countersignature can be verified by `brew` (through a Sigstore client) using the same basic client-side signature verification process described in the task above, making this a progressive enhancement to "bottle-only" signatures.
```

joshuagl (commented):
This is great! The PR is missing some context, but [from Slack](https://openssf.slack.com/archives/C034CBLMQ9G/p1689609210801249) I infer that this is a proposal for Alpha-Omega funding that is sponsored by the WG Securing Software Repos. Is that correct?

di:
> The PR is missing some context, but [from Slack](https://openssf.slack.com/archives/C034CBLMQ9G/p1689609210801249) I infer that this is a proposal for Alpha-Omega funding that is sponsored by the WG Securing Software Repos. Is that correct?

This is just a general proposal from this WG that could be submitted to any potential funder. Although we do plan to submit this to A/O for funding first, there isn't anything A/O-specific here.

steiza (approved):
Thanks for putting this together! I support this proposal.

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
di: 3bfdf07

proposals/code-signing-for-homebrew.md
di: ca6f4a7

woodruffw:
> but [from Slack](https://openssf.slack.com/archives/C034CBLMQ9G/p1689609210801249) I infer that this is a proposal for Alpha-Omega funding that is sponsored by the WG Securing Software Repos. Is that correct?

Correct!

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: > Could we not use cosign style blob signing like they do for wasm and other blobby things?

That's the plan! The "blob" in this case is the attestation payload; my apologies if "payload" was misleading.

(More specifically, my thought here was to format the attestation as an `in-toto` attestation document, reusing/adapting the existing NPM format if possible, and then doing "blob" signing over that. That might need some slight tweaking e.g. to ensure that we submit `intoto` rather than `hashedrekord` to Rekor, but that's the broad idea.)

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,101 @@
...
+2. The signing step is responsible for producing a Sigstore signature that _attests_ to the bottle. This should include, at minimum, the bottle's SHA256 digest and the bottle's filename. The digest is necessary to produce a proof that the signature is for a particular bottle's contents, and the filename is necessary to produce a proof that the signature is intended to accompany a particular install step (i.e. `brew install foo` should not verify if an attacker substitutes `foo`'s bottle for a correctly signed bottle of bar, or even a different version of foo than the resolver produces).
+3. The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a consistent and publicly derivable manner from just the formula and bottle's own metadata. In other words: a client that knows how to access the bottle's resource should also be able to access the bottle's signature resource with no additional metadata.
```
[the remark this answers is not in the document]
woodruffw: Yes! I actually wasn't aware of `cosign triangulate`, but this looks like it will be a helpful reference for our discovery step here (esp. if we end up using a dedicated attestation storage API as mentioned by @steiza)

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
di: I think @lukehinds was asking "why not just sign over the blob's bytes" vs. signing over some attestation payload, and the question at the root is probably: "why does the filename need to be included in the attestation?"

proposals/code-signing-for-homebrew.md
```diff
@@ -0,0 +1,102 @@
...
```
di, line 29:
```suggestion
3. The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a consistent and publicly derivable manner from just the formula and bottle's own metadata. In other words: a client that knows how to access the bottle's resource should also be able to access the bottle's signature resource with no additional metadata, similar to how [signature location and management](https://docs.sigstore.dev/cosign/signing_with_containers/#signature-location-and-management) works for signing container images with Sigstore.
```

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
di: 84124b2410d6074d9e8360ba35b402839bc8c898

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: Oh, gotcha. The reason for signing over the filename as well is to prevent (unintended) domain separation: two potential attack scenarios here:

1. A "mix and match" where an attacker substitutes one correctly signed package version for another (older, vulnerable) one. Without the filename (which includes the version) our signature check would accept that _any_ previously signed package or package version.
2. Cache poisoning or confusion attacks where an attacker manages to confuse Homebrew into signing for and/or serving unintended bottles; signatures over these bottles would be rejected so long as the attestation includes the filename, since they wouldn't match the user's expected bottle.

joshuagl (approved):
> > The PR is missing some context, but [from Slack](https://openssf.slack.com/archives/C034CBLMQ9G/p1689609210801249) I infer that this is a proposal for Alpha-Omega funding that is sponsored by the WG Securing Software Repos. Is that correct?
> 
> This is just a general proposal from this WG that could be submitted to any potential funder. Although we do plan to submit this to A/O for funding first, there isn't anything A/O-specific here.

Got it. The part I was missing from the PR description was that this is a "proposal from this WG that could be submitted to any potential funder".

di:
> Got it. The part I was missing from the PR description was that this is a "proposal from this WG that could be submitted to any potential funder".

I've updated the description!

woodruffw (approved):

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
TomHennen: Have you considered using [a VSA](https://slsa.dev/verification_summary/v1) as the attestation type?  The filename to be checked would go in `resourceUri`, you could use `verifiedLevels` to distinguish between things that have been newly built vs those that have been backfilled.  Note that VSA is likely [moving out of the SLSA repo](https://github.com/slsa-framework/slsa/issues/921#issuecomment-1638535179) and into the in-toto repo to be made more generic.

proposals/code-signing-for-homebrew.md
TomHennen: If backfill attestations are provided for all the bottles would a flag day be required?  E.g. if all bottles have some attestation then the next release of the client can just assume they'll always be there?

woodruffw: > Presumably we can add the flag day logic to clients first, then spend time working on producing signing/attestations, and thus cover a larger group of in-use client versions more quickly when flag day comes.

That's what I was thinking, although @TomHennen's point is also good -- Homebrew tends to "roll" updates very rapidly (i.e., all users rapidly update to the latest stable once released), so with the "backfill" approach there's IMO less value to the flag day.

That being said, I initially worked that in _in case_ we do end up needing it, so IMO it's worth keeping in the proposal as a compatibility/rollout risk mitigation. But we may end up not needing it!

proposals/code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: I might be misunderstanding, but is VSA meant for SLSA provenance _generators_ or for _verifiers_ of that provenance? From that link, it looks like the latter (since it's a countersignature/counterattestation by another party), whereas this proposal is focusing on the former (generating the initial attestation of provenance).

If the above is correct, then I _think_ we use a non-VSA predicate type for the attestations that Homebrew itself is producing, but may want to use the VSA type for the "source attestation" idea mentioned later in the proposal (where Homebrew is verifying a downstream project's source artifact before producing its own build provenance).

(Apologies if I'm way off base in my understanding here!)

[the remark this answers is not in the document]
TomHennen: IMO the VSA can be used anytime you want a downstream user to be able to trust some upstream system's opinion on the 'goodness' of an artifact.  That upstream system could be the same as the one generating the provenance or it could be different.  Either way the VSA is a convenient abstraction for downstream users that don't actually want to evaluate the entire provenance (similar to what code-signing does but with slightly more metadata).

For safety reasons I'd typically expect the system creating the SLSA provenance to be separate from the one doing code-signing or VSA production, but it doesn't have to be.

Using a VSA could also be a straightforward mechanism to resolve point 1 from [your comment here](https://github.com/ossf/wg-securing-software-repos/pull/20#issuecomment-1640397166).  VSAs could be issued for all those old things, either making _no_ statement in `verifiedLevels` or explicitly adding something like `SLSA_BUILD_LEVEL_UNEVALUATED` to make it clear it's not saying anything about how it was built, but still providing the basic protections of code signing.  This is somewhat similar to [the VSA use case described here](https://security.googleblog.com/2022/04/how-to-slsa-part-1-basics.html#:~:text=of%20different%20times.-,On%20import%20to%20repo,-Squirrel%20plans%20to) and [here](https://security.googleblog.com/2022/04/how-to-slsa-part-3-putting-it-all.html#:~:text=Downstream%20users%20of%20this%20Acme%20container%20can%20then%20check%20the%20Acme%20issued%20VSA).

proposals/code-signing-for-homebrew.md
woodruffw: > I suppose one issue is that it's not yet clear how sigstore signed attestations work in bundles?

There's some unfortunate terminology overlap here: the "bundles" in this proposal are Sigstore bundles, which include the signature, signing certificate, transparency log entries, etc. 

The signature in turn is performed over the (in-toto formatted) attestation, which is also committed to the transparency log (since Rekor has a dedicated type for in-toto attestations). 

In other words: the top level "unit" here is a Sigstore bundle, which then contains the signed "thing". That "thing" could itself be the bottle attestation (what I was planning) or an in-toto bundle, but that latter one then means multiple layers of signing and I need to think about that some more 🙂

TomHennen: Yeah, avoiding multiple levels of signing would be good.  FWIW @haydentherapper and I had discussed that it might be worthwhile to figure out either how to

1. Put a _Sigstore_ bundle in an in-toto bundle (e.g. can it just be dropped in a jsonl file and work for folks that expect it), or
2. Figure out how to get Sigstore to embed the signing information it needs inside the DSSE itself instead of wrapping it.

My intuition is that 1 should be pretty easy, but that 2 is the 'right' solution but will take much longer and won't be backwards compatible.

woodruffw: Yeah, (1) seems easier!

I'm a little bit less familiar with in-toto bundles, but from a quick scan they're JSON lines, with each JSON payload being an in-toto `Envelope`. Those envelopes are then signed, so would this be double-signing again? Or am I missing something?

TomHennen: Re @woodruffw 

There doesn't need to be double-signing, what I'm thinking for option 1 is that we can potentially rely on ["Consumers MUST ignore unrecognized lines"](https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md#:~:text=Consumers%20MUST%20ignore%20unrecognized%20lines.) and just format each _Sigstore_ bundle into a single line and add to the _in-toto_ bundle without doing any additional work.  So, a bit of a hack.

Re @adityasaky yes! That seems like the ideal solution (corresponding to (2) above?).  Does anyone think that can happen quickly enough to rely on it for this use case or if we should do (1) instead?

adityasaky: I can submit a proposal to the DSSE spec based on the conversations in that issue but I'm not 100% certain about how its adoption will progress in sigstore. @haydentherapper thoughts?

haydentherapper: https://github.com/sigstore/sig-clients/issues/8 to discuss with the sigstore community a bundle format

TomHennen: > Is the idea future consumers might understanding it, and could do more useful things with it?

Yes exactly.  Consumers would need to support multiple different types of lines in the in-toto bundles, but that seems doable (even if it's inelegant).

di:
Hi folks, thanks for all the reviews to date. Just an FYI that I'd like to merge this by EOD today, so we can submit it to A/O by the end of the week.

haydentherapper:
@di @woodruffw What do you think about adding a threat model as mentioned in https://github.com/ossf/wg-securing-software-repos/pull/20#issuecomment-1638574892?

woodruffw:
> What do you think about adding a threat model as mentioned in [#20 (comment)](https://github.com/ossf/wg-securing-software-repos/pull/20#issuecomment-1638574892)?

I'm happy to do that; I'll open a PR in a bit.

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+
+Authors: [William Woodruff](https://github.com/woodruffw), [Dustin Ingram](https://github.com/di)
+
+## Summary
+
+The document contains a technical proposal for introducing build provenance and cryptographic signatures to the Homebrew package manager. The end goal is to improve Homebrew's authenticity and supply-chain integrity properties, establishing that a given package was both built from an attested source artifact as well as on an attested machine (i.e., a Homebrew-controlled GitHub Actions workflow).
+
+The proposal is broken up into individual technical items, which are meant to be semi-independent units of work. The units of work are ordered by dependency relationship; i.e. signature generation must come before any signatures can be verified on the client side.
+
+## Technical Background
```
trishankatdatadog, line 11: Please add a threat model, because without one, it's hard for the unfamiliar reader to see exactly what this proposal solves for.

Is it fair to say that the threat model is that attackers who compromise the object storage (whatever that is, whether a GitHub repo, or an OCI registry) cannot pretend to be the authors of the package, at least w/o compromising GitHub Actions, Sigstore Fulcio, and/or the GitHub repo for the package?

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+
+The high-level flow for bottle signing:
+
+1. Existing bottle-building workflows will gain a signing step, which will only run as a finalization step for bottles that appear on `master` (i.e., not just the temporary bottles that are built as part of the normal PR lifecycle).
+2. The signing step is responsible for producing an attestation with a Sigstore signature that _attests_ to the bottle. This signed attestation should include, at minimum, the bottle archive's SHA256 digest and the bottle's filename. The digest is necessary to produce a proof that the signature is for a particular bottle's contents, and the filename is necessary to produce a proof that the signature is intended to accompany a particular install step (i.e. `brew install foo` should not verify if an attacker substitutes `foo`'s bottle for a correctly signed bottle of `bar`, or even a different version of `foo` than the resolver produces). The attestation will follow the [in-toto format](https://github.com/in-toto/attestation/tree/main/spec/v1) and thus will be consistent with [SLSA's attestation model](https://slsa.dev/attestation-model).
```
trishankatdatadog, line 30: Please clarify that you are specifically using the in-toto provenance attestation. Otherwise, it's not clear what you're attesting to (e.g., human code reviews, test results).

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+3. The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a predictable and publicly derivable manner from just the formula and bottle's own metadata, [consistent with SLSA's guidenlines on distributing provenance](https://slsa.dev/spec/v1.0/distributing-provenance). In other words: a client that knows how to access the bottle's resource should also be able to access the bottle's signature resource with no additional metadata, similar to how [signature location and management](https://docs.sigstore.dev/cosign/signing_with_containers/#signature-location-and-management) works for signing container images with Sigstore.
+
+At the end of this flow, both the bottle and its attestation with signature should appear in the tap's associated object storage.
```
trishankatdatadog, line 33: Could you clarify where this object storage would be exactly?

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+
+Once a tap has been fully signed for (i.e., all referenced bottles have either a "backfill" or build attestation), the tap must establish a piece of "flag day" metadata: the tap should designate a timestamp `T`, after which all bottles defined by formulae in the tap are signed. This fundamentally requires that the tap itself is honest about its flag day: an attacker who can replace the tap's flag day can perform a downgrade by disabling signature verification. This is no different in scope from Homebrew's existing threat model, where an attacker must not be allowed to modify files in a tap. In other words: this signing scheme only protects the bottles themselves, not the tap's repository; the tap's repository must continue to be protected as a matter of policy.
+
+### Sigstore signatures for third-party taps
+
+While not in scope for this proposal, this work would also enable signing for third-party taps, such as those maintained by independent open source organizations or companies.
+
+For third-party taps hosted on GitHub, the signing flow would be identical: any bottle-building workflow (including reusing the official one) would sign for bottles.
```
trishankatdatadog, line 43: Is it explained in verification below how you would find signatures for first- vs third-party packages?

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+
+And at `brew install` time:
+
+- For each bottle resolved during a `brew install` invocation, the bottle must have an accompanying signature. If the accompanying signature is not present, then `brew install` must fail.
+- For each bottle and signature, `brew install` must internally reconstruct the _attestation_ described in the signing process above. This is then used as the cleartext input for signature verification. This input is then passed into the underlying Sigstore client along with its associated signature; the Sigstore client is responsible for performing both the signature verification and its associated identity verification.
+- If the underlying signature verification fails, then `brew install` must fail.
+
+Like with previous Homebrew features, an experimental feature-flagged variant of signature verification may be merged and enabled before it is enabled by default. Verification will be optional while bottles are incrementally updated/rebuilt with signatures, and once every core bottle is signed, then everything must be signed after the flag day.
+
+## Full source and build provenance for Homebrew
```
trishankatdatadog, line 70: This is a good idea, but should be left to future work IMHO. The Attestation framework is still discussing whether to create a separate [Source attestation](https://github.com/in-toto/attestation/issues/124).

trishankatdatadog: Also, what's not clear to me is how you would enforce and update something like an in-toto layout with policies:

[https://github.com/in-toto/ITE/pull/49](https://www.google.com/url?q=https://github.com/in-toto/ITE/pull/49&sa=D&source=docs&ust=1689883543495248&usg=AOvVaw2Mn7ebOL7Hn9gph6EZme64)

[https://github.com/in-toto/ITE/pull/50](https://www.google.com/url?q=https://github.com/in-toto/ITE/pull/50&sa=D&source=docs&ust=1689883543495396&usg=AOvVaw0KzVp7AcJq1CwexF8bmGc2)

A simpler alternative is to hardcode the verification in the brew binary, although this is not recommended for many good reasons.

Cc @adityasaky

trishankatdatadog: Last but not least, what's missing here is a [publish attestation](https://github.com/npm/attestation/tree/main/specs/publish/v0.1) if the object storage is indeed different from GitHub. This will help catch MitM attacks between GitHub Actions and the object storage.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: https://github.com/ossf/wg-securing-software-repos/pull/25

woodruffw: > Is it fair to say that the threat model is that attackers who compromise the object storage (whatever that is, whether a GitHub repo, or an OCI registry) cannot pretend to be the authors of the package, at least w/o compromising GitHub Actions, Sigstore Fulcio, and/or the GitHub repo for the package?

Correct: this is scenario (1) in the PR I linked above.

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+
+The high-level flow for bottle signature verification, at tap time:
+
+- A user performs either an explicit tap (`brew tap foo/bar`) or an implicit tap (`brew install foo/bar/baz`), and the local `brew` client establishes a trust proposition based on that action ("bottles from `foo/bar` must be signed by the `foo/bar` identity").
+- The local `brew` client looks for the "flag day" metadata and, if present and before the local system time, enforces signature verification for all bottles referenced by the tap.
+
+And at `brew install` time:
+
+- For each bottle resolved during a `brew install` invocation, the bottle must have an accompanying signature. If the accompanying signature is not present, then `brew install` must fail.
+- For each bottle and signature, `brew install` must internally reconstruct the _attestation_ described in the signing process above. This is then used as the cleartext input for signature verification. This input is then passed into the underlying Sigstore client along with its associated signature; the Sigstore client is responsible for performing both the signature verification and its associated identity verification.
```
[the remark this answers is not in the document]
adityasaky: Also, the signature is calculated across the attestation which records a hash of the artifact being installed. So in addition to verifying the signature, the client must ensure the attestation's claim re the artifact's hash matches the hash of the received artifact.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: That partially depends on @steiza 🙂 

The answer here is not particularly satisfying: third-party taps can put their object storage wherever they please, and Homebrew isn't prescriptive about that (other than having support for a few download strategies already built in).

The "simpler" answer is that this proposal is initially targeted at `homebrew-core`, and `homebrew-core` *currently* repurposes GitHub Packages to do object storage for its bottles. However, that's an implementation detail, so I'd prefer to leave that out (in the interest of keeping the proposal mostly behavioral in structure). 

That being said: if you think it'll be helpful to add a note about *specific* implementation possibilities, I can do that!

proposals/code-signing-for-homebrew.md
woodruffw: Makes sense to me. I'm happy to bundle the actual signing outputs here in whatever format makes things easiest for additional future users; if that means rolling it into another surrounding bundle, that's fine!

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: I think that's covered by the language above:

> The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a predictable and publicly derivable manner from just the formula and bottle's own metadata

In other words: when you do `brew install foo` with Homebrew, `brew` is responsible for consulting the tap metadata corresponding to `foo` and determining the appropriate download strategy for that tap.

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,112 @@
...
+- If the underlying signature verification fails, then `brew install` must fail.
+
+Like with previous Homebrew features, an experimental feature-flagged variant of signature verification may be merged and enabled before it is enabled by default. Verification will be optional while bottles are incrementally updated/rebuilt with signatures, and once every core bottle is signed, then everything must be signed after the flag day.
+
+## Full source and build provenance for Homebrew
+
+Bottle signing and subsequent verification unlocks a potential subsequent improvement: because Homebrew performs source builds to produce bottles, Homebrew could also verify signatures for those source artifacts. That signature could then be _countersigned_ for by Homebrew's own bottle signatures, effectively producing a strong cryptographic association from the original source artifact (produced by a third-party upstream) to the Homebrew-produced binary artifact.
```
adityasaky, line 72: The idea is that the source (say the tarball of ripgrep from the example below) is signed by its author and brew install would verify its signature as well as the provenance for the build signed by homebrew-core's identity.

I'm trying to understand the value of the signature on the source tarball. In option 1, the signature is expected to be from the identity associated with the github repo. Is the ability to issue this signature for the tarball more protected than the capability to upload a release? Does it require 2FA every time, for example?

Option 2 is more interesting because the upstream maintainer for a package can be associated with a separate, non-github identity, but I'm also curious about the overhead involved. Does it fall to the homebrew maintainers to verify that the github repo and the secondary ID are owned by the same person? Also, how is the other identity updated / rotated / revoked? Right now it seems like all of this is manual.

Finally, apart from verifying a signature on the source and verifying the built artifact's provenance / sig, the two must also be cross checked to ensure the source signed in the first case is identical to what's used as the input to the build. As @trishankatdatadog notes above, in-toto layouts help here.

mnm678: I'll add that TUF would be a good option for doing the association for option 2. An offline delegation to a separate developer identity would provide additional security, and provide an easy way for clients to determine which identities to use for verification.

haydentherapper: Note that developer identity is out of scope currently for the MVP, that the identity provided by GitHub Actions and is linked to source will be used.

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: > Also, the signature is calculated across the attestation which records a hash of the artifact being installed. So in addition to verifying the signature, the client must ensure the attestation's claim re the artifact's hash matches the hash of the received artifact.

Right, that's what I meant by "reconstruct": the verifying party has to have an independent reason to trust the attestation, which can be either "the hash + other materials in the attestation are what I think they should be" or "I ignored the supplied attestation and re-built it entirely from independent state, and it matches."

The latter is less prone to false-positive verifications due to format changes: if the attestation is always reconstructed from scratch, the verifier can't accidentally forget to cross-check a newly mandatory field.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: Thanks; I'll roll this into #25.

woodruffw: 2968c47

proposals/build-provenance-and-code-signing-for-homebrew.md
adityasaky: @haydentherapper, so the signature on the source tarball must be issued by github actions on that repository? As in, the release must be automated using github actions?

proposals/build-provenance-and-code-signing-for-homebrew.md
trishankatdatadog: Agreed: the exact choice of object storage doesn't matter, so long as the threat model covers that, no matter what, it can be compromised and tampered with.

proposals/build-provenance-and-code-signing-for-homebrew.md
haydentherapper: Correct, given all core bottles are currently built using GitHub Actions, we can construct an identity that represents the GitHub Action. https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md, in particular https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#1361415726419--build-signer-uri, has more information.

See https://github.com/ossf/wg-securing-software-repos/commit/0f9675a427973d84ee56f84c821b12dd526c987c

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
adityasaky: The attestation is likely to include information that the installer may not know about (because it does not care about it) if the idea is to use the existing provenance spec. For example, provenance will record information about the artifacts used as inputs to the build. Setting aside source signing, the client has no need to know the list of input artifacts AND all their hashes. I agree that reconstruction is possible for sticky expected fields like `buildDefinition.buildType == <expected value>`.

Instead, IMO the client should:
1. Download artifact + attestation
2. Verify attestation's signature using homebrew-core's gha identity
3. Verify attestation's subject matches artifact by digest
4. Verify other expected fields that are _currently_ relevant

IMO optimizing for scenarios where a new field is added and so on will mean severely constraining the information actually recorded in the attestations.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: Yeah, this section was written with machine identity in mind: it assumes something like GitHub Actions (or another build/CI identity supported by Sigstore), which means that verifying and then countersigning the upstream source's signature here would include embedded claims about the source provenance (e.g. the source URL, which workflow built it, etc.).

>  the signature is expected to be from the identity associated with the github repo. Is the ability to issue this signature for the tarball more protected than the capability to upload a release? Does it require 2FA every time, for example?

The ability to create a verifiable machine identity signature on GitHub Actions is gated behind a workflow permission, which is comparable (in practice) to having write access to the repository. So it's *similar* to having the ability to make a release; the value comes from it being verifiably bound to the identity that made the release.

trishankatdatadog: May I then recommend refraining from talking about Source altogether for the MVP? Because I think the problem is trickier to solve for right now than it looks like. We can certainly save it for a future iteration 🙂

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: Cool; let me know if you think the language in #25 covers that!

trishankatdatadog: Yes, I think it's covered by "Model 1: Compromise of bottle storage". Thanks!

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
TomHennen: I think this is why I was suggesting using the VSA as the attestation to verify, the client doesn't need to know very much at all about how the thing was built (and it also allows backfil).

My read is that the proposal suggests two things 1) the provenance 2) a signature over something else

(2) is what would be used by the client, the question is just what the something else should be?

I know https://github.com/ossf/wg-securing-software-repos/commit/2968c479f4251879005ecfddcc7f911aee9fafa3 clarified the thing in (2) to be the provenance, but I don't think that will work well for this use case because the clients will have to know about all these fields to check (complicated) and it can't be backfilled for existing bottles.

Somehow I don't see the VSA discussion anymore in the main view, but it still seems to live on here: https://github.com/ossf/wg-securing-software-repos/pull/20#discussion_r1268338827

If you don't want to use the VSA in particular then the publish attestation npm is using would probably work too; though I think it would be a shame since the VSA would solve this so well. :)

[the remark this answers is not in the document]
trishankatdatadog: I'm sorry, this isn't clear to me either: why would a client _reconstruct_ its own attestation instead of trusting the signed one? Otherwise, why sign the attestations if you don't fully trust it?

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: Works for me -- I think this is technically outside of the scope anyways (since we're focusing on build provenance for this proposal and not source provenance), so if @di is okay with an "out of scope" note I'll add it.

haydentherapper: FWIW, source in this case != source provenance. There's a link between source and build because of the values in the GHA identity. It's accurate to say we're not tackling source provenance, but being able to link the identity to source to build is what makes verification seamless for this.

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
trishankatdatadog: Anyway, maybe we're just misunderstanding each other here. We can certainly work out the full details of verification later. I suggest keeping it general for now 🙂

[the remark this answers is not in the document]
adityasaky: @TomHennen is the idea to generate provenance, then immediately verify and generate a VSA? I'm wary of using VSA's directly to communicate data (perhaps SCAI is a better fit there) because VSA implies a higher level of verification. IMO there are two things here, yes the client doesn't need to know much but if there isn't a resource constraint, it's probably better to stick to provenance. The reason to _compulsorily_ use VSA is if the client _must_ reconstruct the attestation, as VSA has more straightforward claims, the client should be able to do so.

proposals/build-provenance-and-code-signing-for-homebrew.md
trishankatdatadog: So, if I understand @haydentherapper and @woodruffw correctly, the idea is that the _provenance_ attestation cryptographically binds the source code repository (and the build platform) to the artifact, correct?

haydentherapper: Correct, nothing about the legitimacy of the source code. Concretely, here's an example of an identity: https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@v1.4.0

This includes the source code repository and the versioned builder (reusable workflow) that creates the artifact.

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: @adityasaky:

> IMO optimizing for scenarios where a new field is added and so on will mean severely constraining the information actually recorded in the attestations.

This is potentially a controversial opinion of mine, but I think this is good! Malleability during verification gets treacherous very quickly: the fact that some fields are checked some of the time introduces state that an attacker can influence. 

Those kinds of risks can be offset by strong policy controls, but IMO the complexity of this *particular* case doesn't warrant that kind of flexibility: all we _actually_ want to put in the attestation here is the bottle's digest and its filename. Everything else (identity, build provenance) comes for "free" from the Sigstore side, thanks to the machine identities we're using.

---

@TomHennen:

> I think this is why I was suggesting using the VSA as the attestation to verify, the client doesn't need to know very much at all about how the thing was built (and it also allows backfil).

Yeah, I think I'm understanding what you intended now: per above I think all we _really_ want here is the bottle's digest and filename, so the full in-toto provenance attestation might be too "heavy" for this particular use case.

> Somehow I don't see the VSA discussion anymore in the main view, but it still seems to live on here: https://github.com/ossf/wg-securing-software-repos/pull/20#discussion_r1268338827

Sorry, that got lost in the mix!

If we were to use a VSA-formatted attestation for this, what would that _concretely_ look like? Something like this: https://slsa.dev/verification_summary/v1#example?

If so, I think I'm still a little on which fields correspond to what: is the idea that `resourceUri` would be the filename, and the bottle digest would live somewhere else?

---

@trishankatdatadog:

> I'm sorry, this isn't clear to me either: why would a client _reconstruct_ its own attestation instead of trusting the signed one? Otherwise, why sign the attestations if you don't fully trust it?

The attack I'm thinking of is similar to CVE-2022-23649: unless the verifying party cross-checks the attestation (either by re-building it, or checking every relevant field, which may change over time), a malicious party can provide a _valid but contextually wrong_ attestation, e.g. for a different (correctly signed) bottle.

@adityasaky's point mitigates this: if you check the hash in the attestation, then the bottle's contents are bound to the attestation. But there are other holes: the bottle's _filename_ also needs to be checked to prevent namespace or version confusion, and there may be future fields that we haven't thought of yet that _also_ need to be added and checked.

That's why I generally think of attestation verification as a "reconstruction" problem, rather than a "check what you know" problem.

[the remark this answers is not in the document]
trishankatdatadog: > That's why I generally think of attestation verification as a "reconstruction" problem, rather than a "check what you know" problem.

Not every field in a signed piece of metadata is going to be relevant to what you are looking for. As long as you check every field you _should_ check for what you are looking for (which may or may not be all fields), you should be OK. Anyway, like I said before, I do not recommend that we get hung up on the precise details of verification right now, so long as we can all agree later. For this reason, I recommend removing details and keeping it general for now.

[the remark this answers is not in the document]
woodruffw: Sounds good. In that case, I'm going to do another follow-up to emphasize behavior rather than a specific format in this proposal: I think for the MVP we really just want the bottle's digest and filename, and the format/shape that comes in is TBD with the engineering effort itself.

(I'll also reword to clarify that reconstructing is just one way to establish confidence in an attestation; checking a subset of fields as a matter of policy can also be appropriate, as you've noted.)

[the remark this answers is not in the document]
TomHennen: @adityasaky
> @TomHennen is the idea to generate provenance, then immediately verify and generate a VSA?

I generally assume that lots of builds happen but they don't all get published.  I don't know the specific architecture of Homebrew though.  Maybe that's how it would work here?  I'd actually expect different systems being involved: one (or many) doing builds and producing provenance -> some of those builds being considered for publication -> the ones that do get published get VSAs documenting whatever properties were checked in verification in `verifiedLevels` (which might be none!).

@woodruffw 
> If we were to use a VSA-formatted attestation for this, what would that concretely look like? Something like this: https://slsa.dev/verification_summary/v1#example?

Yeah I think so.
 
> If so, I think I'm still a little on which fields correspond to what: is the idea that resourceUri would be the filename, and the bottle digest would live somewhere else?

The bottle digest would go in `subject.digest` at the top.

When you say filename do you mean the absolute path on disk, or the name of the bottle?  FWIW I'd expect resourceUri to contain the name of the bottle, ideally in purl form.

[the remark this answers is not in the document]
woodruffw: Just the name of the bottle, so thanks!

I guess the only question from there: the VSA format appears to contain its own timestamp, which is marked as required. I *think* we might not want that for bottle signing in Homebrew, since (1) the transparency service is providing our timestamp, and (2) the verifying client doesn't know it ahead of time, meaning that we can't fully reconstruct the attestation during verification. 

But that goes back to the conversation up thread, and I'm going to tweak the proposal to defer that entirely as an engineering decision (since the attestation format here fundamentally doesn't matter for securing Homebrew, so long as it covers the basic data points we need).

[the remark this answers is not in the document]
woodruffw: https://github.com/ossf/wg-securing-software-repos/pull/26

woodruffw (approved):
This is good on my end; thanks to all who reviewed!

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
adityasaky: Thanks! Could you also generalize the text re "reconstruction"? Depending on the attestation format used, reconstruction may not be viable.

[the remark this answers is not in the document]
woodruffw: Sorry, missed that. Doing now.

proposals/code-signing-for-homebrew.md
woodruffw: Resolved per ref'd commit above.

proposals/code-signing-for-homebrew.md
woodruffw: Resolved per #27.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: Resolved per #25.

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
woodruffw: Resolved per #27.

proposals/build-provenance-and-code-signing-for-homebrew.md
woodruffw: I believe this is resolved per convo in https://github.com/ossf/wg-securing-software-repos/pull/20#discussion_r1269928353.

proposals/build-provenance-and-code-signing-for-homebrew.md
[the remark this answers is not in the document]
adityasaky: Thanks!

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Technical Items
+
+### Provenance attestations and signatures for official Homebrew taps
+
+Homebrew's packaging processes are built around GitHub Actions. In particular, Homebrew uses GitHub Actions to perform builds of all core formulae (package specifications, like LLVM, `bash`, or `git`) into bottles (binary distributions that can be installed on client machines, similar to an `.app` or `.pkg`). Bottles are then hosted on GitHub Packages (or other hosts, for third-party taps) and are installed by end-user clients e.g. during `brew install` invocations.
+
+Because Homebrew uses GitHub Actions for all `homebrew-core` builds, the Homebrew CI has trivial access to GitHub Actions' [ambient OpenID Connect identities](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect). These identities can in turn be used to generate build provenance and sign with Sigstore, meaning that Homebrew's existing CI workflows would require relatively little additional configuration (and zero external key management) to produce valid SLSA attestations and Sigstore signatures for each bottle build.
```
trishankatdatadog, line 25:
```suggestion
Because Homebrew uses GitHub Actions for all `homebrew-core` builds, the Homebrew CI has trivial access to GitHub Actions' [ambient OpenID Connect identities](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect). These identities can in turn be used to generate build provenance and sign with Sigstore, meaning that Homebrew's existing CI workflows would require relatively little additional configuration (and zero external key management) to produce valid in-toto attestations and Sigstore signatures for each bottle build.
```

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,174 @@
...
+
+2. The signing step is responsible for producing an attestation with a Sigstore signature that _attests_ to the bottle. This signed attestation should include, at minimum, the bottle archive's SHA256 digest and the bottle's filename. The digest is necessary to produce a proof that the signature is for a particular bottle's contents, and the filename is necessary to produce a proof that the signature is intended to accompany a particular install step (i.e. `brew install foo` should not verify if an attacker substitutes `foo`'s bottle for a correctly signed bottle of `bar`, or even a different version of `foo` than the resolver produces). The exact attestation format is subject to further engineering considerations; current candidates include the the [in-toto provenance attestation format](https://slsa.dev/provenance/v1), the [SLSA VSA attestation format](https://slsa.dev/verification_summary/v1) and an adaptation of the [NPM attestation format](https://github.com/npm/attestation/tree/main/specs/publish/v0.1). Regardless of the format chosen, the final attestation will be consistent with [SLSA's attestation model](https://slsa.dev/attestation-model).
```
trishankatdatadog, line 31:
```suggestion
2. The signing step is responsible for producing an attestation with a Sigstore signature that _attests_ to the bottle. This signed attestation should include, at minimum, the bottle archive's SHA256 digest and the bottle's filename. The digest is necessary to produce a proof that the signature is for a particular bottle's contents, and the filename is necessary to produce a proof that the signature is intended to accompany a particular install step (i.e. `brew install foo` should not verify if an attacker substitutes `foo`'s bottle for a correctly signed bottle of `bar`, or even a different version of `foo` than the resolver produces). The exact attestation format is subject to further engineering considerations; current candidates include the the [in-toto provenance attestation format](https://slsa.dev/provenance/v1) (designed for SLSA v1.0), the [in-toto VSA attestation format](https://slsa.dev/verification_summary/v1) (also designed for SLSA v1.0), and an adaptation of the [NPM attestation format](https://github.com/npm/attestation/tree/main/specs/publish/v0.1). Regardless of the format chosen, the final attestation will be consistent with [SLSA's attestation model](https://slsa.dev/attestation-model).
```

proposals/build-provenance-and-code-signing-for-homebrew.md
```diff
@@ -0,0 +1,174 @@
...
+
+3. The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a predictable and publicly derivable manner from just the formula and bottle's own metadata, [consistent with SLSA's guidenlines on distributing provenance](https://slsa.dev/spec/v1.0/distributing-provenance). In other words: a client that knows how to access the bottle's resource should also be able to access the bottle's signature resource with no additional metadata, similar to how [signature location and management](https://docs.sigstore.dev/cosign/signing_with_containers/#signature-location-and-management) works for signing container images with Sigstore.
```
trishankatdatadog, line 33:
```suggestion
3. The signing step is responsible for uploading the resulting signature to the same object storage as the bottle, and ensuring that the uploaded signature is addressable in a predictable and publicly derivable manner from just the formula and bottle's own metadata, [consistent with SLSA's guidelines on distributing provenance](https://slsa.dev/spec/v1.0/distributing-provenance). In other words: a client that knows how to access the bottle's resource should also be able to access the bottle's signature resource with no additional metadata, similar to how [signature location and management](https://docs.sigstore.dev/cosign/signing_with_containers/#signature-location-and-management) works for signing container images with Sigstore.
```

trishankatdatadog (changes requested):

di:
Thanks everyone for your feedback here!

PR #31, opened 2023-08-01
Update build provenance doc to mention privacy concerns

znewman01:
For reasons that are not very interesting to anybody else, it would be very helpful to me if there were an "official statement" from this group to the effect of "privacy concerns are one issue preventing Sigstore adoption for package signing by authors."

This doc feels like a good candidate to me, as we already mention one reason why we don't use human identities.


docs/build-provenance-for-all-package-registries.md
```diff
@@ -32,7 +32,7 @@ These requirements mean you won't be able to generate trustworthy build provenan
...
 
-Note that **build provenance does not try to tie a package back to a verified human identity**. Instead we're focusing on the workload identity from the build system, as that is what produces the package. Strongly authenticating humans and trying to determine their intent is a complex problem space that is out of scope for build provenance!
+Note that **build provenance does not try to tie a package back to a verified human identity**. Instead we're focusing on the workload identity from the build system, as that is what produces the package. Strongly authenticating humans and trying to determine their intent is a complex problem space that is out of scope for build provenance! Further, authenticating humans raises privacy and legal concerns that pose obstacles to repository adoption for signatures; prioritizing build provenance sidesteps these.
```
steiza, line 35:
```suggestion
Note that **build provenance does not try to tie a package back to a verified human identity**. Instead we're focusing on the workload identity from the build system, as that is what produces the package. Strongly authenticating humans and trying to determine their intent is a complex problem space that is out of scope for build provenance! Further, authenticating humans often involves Personally Identifiable Information, which registries would like to avoid additional handling of for privacy and legal concerns. By using workload identity for build provenance, registries can avoid these issues.
```

steiza (approved):

Issue #16, opened 2023-06-20
Produce whitepaper of recommendations for securing package repositories 


znewman01:
This working group has produced a ton of useful information about how best to build a secure package repository, along with data on what repositories are currently doing. Can we crystallize this into an easy-to-digest guide to package repository security for package repository admins/maintainers? Topics would include (by no means complete):

- list of incidents: [torchtriton](https://pytorch.org/blog/compromised-nightly-dependency/)
- guidance on dependency selection: keeping dependency trees trim 
   -  [trust metrics](https://arxiv.org/abs/2306.00240)
- revocation policies
- encouragin best practices
  - [OSSF Scorecards](https://github.com/ossf/scorecard); [do scorecards help?](https://arxiv.org/abs/2210.14884); [Sonatype safety rating](https://ossindex.sonatype.org/doc/sonatype-safety-rating)
- static/dynamic analysis
  - malware scanning: [DataDog/guarddog](https://github.com/DataDog/guarddog); [benchmark; does it help?](https://arxiv.org/abs/2209.13288); https://github.com/ossf/wg-securing-software-repos/issues/8
  - secret scanning ([GitHub](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning), [GitLab](https://docs.gitlab.com/ee/user/application_security/secret_detection/))
  - reporting APIs
- security auditing
  - for packages (attestations?)
  - for the repository itself
- package namespacing
- dependency confusion
  - [Python discussion](https://discuss.python.org/t/proposal-preventing-dependency-confusion-attacks-with-the-map-file/23414)
- package signing/attestations
  - https://github.com/ossf/wg-securing-software-repos/issues/10
  - [TUF](https://theupdateframework.com/), [in-toto](https://in-toto.io/)
  - [npm](https://github.blog/2022-08-08-new-request-for-comments-on-improving-npm-security-with-sigstore-is-now-open/)/pypi as case studies
  - [SLSA](https://slsa.dev/): provenance
  - [publish attestations](https://github.com/npm/attestation/tree/main/specs/publish/v0.1), [Docker attestations](
- https://docs.docker.com/build/attestations/)
  - with [Sigstore](https://www.sigstore.dev/)?
  - privacy concerns
- publishing workflows ([PyPI trusted publishers](
- https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/))
  - API token scoping
- account management
[...]

PR #37, opened 2023-12-13
Add Principles for Package Repository Security

steiza:
For https://github.com/ossf/wg-securing-software-repos/issues/16

A security maturity model for package repositories, for assessing current capabilities and roadmapping future improvements.

### DIFF
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -43,7 +43,7 @@ Applies to: package repositories that have user accounts.
 - Level 1
   - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
   - The package repository documents their account recovery policy.
-  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
+  - The package repository supports strong multi-factor authentication (MFA) via (at minimum) TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
   - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
   - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
 - Level 2
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -75,7 +75,7 @@ Applies to: package repositories that have user accounts and accept built packag
 Applies to: all package repositories
 
 - Level 1
-  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities to the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
   - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
 - Level 2
   - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -62,6 +62,7 @@ Applies to: package repositories that have user accounts and accept built packag
 
 - Level 1
   - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
+  - API tokens are prefixed with a repository-specific identifier
 - Level 2
   - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
 - Level 3
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -82,7 +82,7 @@ Applies to: all package repositories
   - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
   - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
 - Level 3
-  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
+  - Undergo periodic security reviews of package repository infrastructure (and optionally publish remediations once they are complete)
   - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
   - Advisories for malicious packages are published using the OSV schema and aggregated by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages))
 
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -5,13 +5,13 @@ Last updated: Jan 2024
 
 ## Background
 
-The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
+The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to offer a set of best practices that package repositories should strive to adhere to.
 
-By package repository (sometimes also called package registry) we're referring to the service that stores and index packages, that users and clients can download packages from. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
+By package repository (sometimes also called package registry) we're referring to the service that stores and indexes packages, enabling users and clients to download packages. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
 
-The taxonomy is because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
+We include the below taxonomy because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
 
-The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)) or write specific grant proposals that reference this guidance.
+The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)), or write specific grant proposals that reference this guidance.
 
 ## Taxonomy of Package Repositories
 
@@ -32,9 +32,9 @@ Does the package repository ...
 
 We define 4 levels of security maturity:
 - Level 0 is defined as having very little security maturity.
-- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication and allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
-- Level 2 is defined as having moderate security.
-- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
+- Level 1 is defined as having basic security maturity, which includes supporting basic security features like multi-factor authentication (MFA) and allowing security researchers to report vulnerabilities. All package management ecosystems should be working towards at least this level.
+- Level 2 is defined as having moderate security, which includes actions like requiring MFA for critical packages and warning users of known security vulnerabilities.
+- Level 3 is defined as having advanced security, which includes actions like requiring MFA for all maintainers and supporting build provenance for packages. This level is more aspirational, especially for smaller package management ecosystems.
 
 ### Authentication
 
@@ -43,18 +43,18 @@ Applies to: package repositories that have user accounts.
 - Level 1
   - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
   - The package repository documents their account recovery policy.
-  - The package repository supports strong multi-factor authentication (MFA) via (at minimum) TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
+  - The package repository supports strong multi-factor authentication (MFA) via, at minimum, TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
   - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
   - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
 - Level 2
   - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and removing the ability to recover an account via an email domain that has been abandoned.
-  - The package repository supports MFA via phishing-resistant means such as WebAuthn
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn.
   - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
   - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using known leaked credentials upon login. If a user has a leaked credential, detected upon login, the package repository prompts them to change it. The package manager checks passwords at registration and prevents users from registering with known leaked credentials.
 - Level 3
   - The package repository supports passwordless authentication, such as passkeys.
   - The package repository requires MFA for all maintainers.
-  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g. top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
+  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g., the top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
 
 ### Authorization
 
@@ -62,12 +62,12 @@ Applies to: package repositories that have user accounts and accept built packag
 
 - Level 1
   - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
-  - API tokens are prefixed with a repository-specific identifier
+  - API tokens are prefixed with a repository-specific identifier.
 - Level 2
   - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
 - Level 3
-  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
-  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false-positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
+  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/).
+  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
   - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
 
 ### General Capabilities
@@ -75,16 +75,17 @@ Applies to: package repositories that have user accounts and accept built packag
 Applies to: all package repositories
 
 - Level 1
-  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to identify and report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so within the requirements of the policy. The package repository should strive to remediate reported vulnerabilities within 90 days.
   - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
 - Level 2
   - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
   - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
   - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
+  - The package repository warns of known security vulnerabilities in dependencies in the package repository UI.
 - Level 3
   - Undergo periodic security reviews of package repository infrastructure (and optionally publish remediations once they are complete)
   - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
-  - Advisories for malicious packages are published using the OSV schema and aggregated by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages))
+  - Advisories for malicious packages are published using a standardized machine readable format, such as using the OSV schema and aggregation provided by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages)).
 
 ### CLI Tooling
 
@@ -93,10 +94,11 @@ Applies to: all package management ecosystems, although here we are primarily fo
 - Level 1
   - The CLI allows installing dependencies that are pinned based on hash, version, etc.
 - Level 2
-  - The package repository warns of known security vulnerabilities in dependencies, in both the package repository UI and CLI tools when installing.
+  - The CLI warns of known security vulnerabilities in dependencies when installing packages.
 - Level 3
   - The CLI has functionality to produce SBOMs.
-  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln))
+  - The CLI has functionality to, where possible, automatically remediate known vulnerabilities in dependencies by upgrading them.
+  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln)).
 
 ## Resources / Inspiration
 
@@ -104,3 +106,5 @@ Applies to: all package management ecosystems, although here we are primarily fo
 - <https://github.com/ossf/wg-securing-software-repos/blob/main/survey/2022/results.csv>
 - <https://docs.npmjs.com/threats-and-mitigations>
 - <https://www.microsoft.com/en-us/research/uploads/prod/2018/03/build-systems.pdf> (as a framework for comparing options in a problem space)
+
+Note: CISA does not endorse any commercial entity, product, company, or service, including any entities, products, or services linked within this document. Any reference to specific commercial entities, products, processes, or services by service mark, trademark, manufacturer, or otherwise, does not constitute or imply endorsement, recommendation, or favoring by CISA.
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -32,7 +32,7 @@ Does the package repository ...
 
 We define 4 levels of security maturity:
 - Level 0 is defined as having very little security maturity.
-- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication and allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
 - Level 2 is defined as having moderate security.
 - Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
 
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -67,7 +67,7 @@ Applies to: package repositories that have user accounts and accept built packag
   - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
 - Level 3
   - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
-  - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
+  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false-positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
   - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
 
 ### General Capabilities
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -1,0 +1,105 @@
+# Principles for Package Repository Security
+
+Authors: [Jack Cable](https://github.com/cablej), [Zach Steindler](https://github.com/steiza)
+Last updated: Dec 2023
+
+## Background
+
+The Securing Software Repositories Working Group (WG) of the OpenSSF is putting together a taxonomy of package repositories and a set of principles for security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
+
+By package repository (sometimes also called package registry) we're referring to the service that stores and index packages, that users and clients can download packages from. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
+
+The taxonomy is because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
+
+The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)) or write specific grant proposals that reference this guidance.
+
+## Taxonomy of Package Repositories
+
+Security capabilities will differ based on the services that the package repository offers. For instance, if the package repository has user accounts, the package repository will need to enforce authentication securely. In this section, we lay out the various relevant aspects of package repositories.
+
+Does the package repository ...
+
+... have user accounts?
+- Yes: then you have to manage things like authentication and account recovery. For example: PyPI
+- No: For example: {index, proxy, sum}.golang.org
+
+... accept built packages, do builds on behalf of users, or only host source code?
+- Accept built packages: then you have to provide a way to link packages back to source code and build instructions. For example: npm
+- Builds on behalf of users: For example, Homebrew
+- Only host source code: For example: {index, proxy, sum}.golang.org
+
+## Security Capabilities of Package Repositories
+
+We define 4 levels of security maturity:
+- Level 0 is defined as having very little security maturity.
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
+- Level 2 is defined as having moderate security.
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and locking accounts for which the domain is abandoned.
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn
+  - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using leaked credentials. If a user has a leaked credential, the package repository prompts them to change it upon login.
+- Level 3
+  - The package repository supports passwordless authentication, such as passkeys.
+  - The package repository requires MFA for all maintainers.
+  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g. top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
+
+### Authorization
+
+Applies to: package repositories that have user accounts and accept built packages.
+
+- Level 1
+  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
+- Level 2
+  - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
+- Level 3
+  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
+  - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
+  - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
+
+### General Capabilities
+
+Applies to: all package repositories
+
+- Level 1
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities to the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
+  - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
+  - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
+  - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
+- Level 3
+  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
+  - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
+  - Malicious packages are published using OSV (e.g. [here](https://github.com/ossf/malicious-packages))
+
+### CLI Tooling
+
+Applies to: all package management ecosystems, although here we are primarily focused on the CLI tooling instead of the registry itself.
+
+- Level 1
+  - The CLI allows installing dependencies that are pinned based on hash, version, etc.
+- Level 2
+  - The package repository warns of known security vulnerabilities in dependencies, in both the package repository UI and CLI tools when installing.
+- Level 3
+  - The CLI has functionality to produce SBOMs.
+  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln))
+
+## Resources / Inspiration
+
+- <https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md>
+- <https://github.com/ossf/wg-securing-software-repos/blob/main/survey/2022/results.csv>
+- <https://docs.npmjs.com/threats-and-mitigations>
+- <https://www.microsoft.com/en-us/research/uploads/prod/2018/03/build-systems.pdf> (as a framework for comparing options in a problem space)
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -1,7 +1,7 @@
 # Principles for Package Repository Security
 
-Authors: [Jack Cable](https://github.com/cablej), [Zach Steindler](https://github.com/steiza)
-Last updated: Dec 2023
+Authors: [Jack Cable (CISA)](https://github.com/cablej), [Zach Steindler](https://github.com/steiza)
+Last updated: Jan 2024
 
 ## Background
 
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -5,7 +5,7 @@ Last updated: Dec 2023
 
 ## Background
 
-The Securing Software Repositories Working Group (WG) of the OpenSSF is putting together a taxonomy of package repositories and a set of principles for security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
+The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
 
 By package repository (sometimes also called package registry) we're referring to the service that stores and index packages, that users and clients can download packages from. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
 
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -84,7 +84,7 @@ Applies to: all package repositories
 - Level 3
   - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
   - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
-  - Malicious packages are published using OSV (e.g. [here](https://github.com/ossf/malicious-packages))
+  - Advisories for malicious packages are published using the OSV schema and aggregated by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages))
 
 ### CLI Tooling
 
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -50,7 +50,7 @@ Applies to: package repositories that have user accounts.
   - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and removing the ability to recover an account via an email domain that has been abandoned.
   - The package repository supports MFA via phishing-resistant means such as WebAuthn
   - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
-  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using leaked credentials. If a user has a leaked credential, the package repository prompts them to change it upon login.
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using known leaked credentials upon login. If a user has a leaked credential, detected upon login, the package repository prompts them to change it. The package manager checks passwords at registration and prevents users from registering with known leaked credentials.
 - Level 3
   - The package repository supports passwordless authentication, such as passkeys.
   - The package repository requires MFA for all maintainers.
--- a/docs/principles-for-package-repository-security.md
+++ b/docs/principles-for-package-repository-security.md
@@ -65,7 +65,7 @@ Applies to: package repositories that have user accounts and accept built packag
 - Level 2
   - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
 - Level 3
-  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
+  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
   - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
   - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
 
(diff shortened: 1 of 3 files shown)


Vombato (commented):
For what's worth, it looks good to me 😁

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and locking accounts for which the domain is abandoned.
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn
+  - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using leaked credentials. If a user has a leaked credential, the package repository prompts them to change it upon login.
+- Level 3
+  - The package repository supports passwordless authentication, such as passkeys.
+  - The package repository requires MFA for all maintainers.
```
the8472, line 56: I think blanket requirements introduce friction and may be onerous for casual/new developers that want to publish hobby projects and such.

Instead individual packages should be labeled based on their level so that consumers who require higher levels can restrict them without burdening those who do not require this.

Generally the commercial/big corporate/government users shouldn't shape registries solely to their needs.
Being able to filter the content based on the criteria they're interested in should provide the same outcome.

ljharb: Authenticator apps are “proper” just like hardware keys, so i don’t think this is much of a consideration.

the8472: Authenticator apps may sit on the same machine as the password manager or on devices that are generally more compromisable. So they provide a lower security level than dedicated hardware tokens.

AIUI webauthn also supports attestation, which means an authentication provider may choose to lock MFA to a set of trusted (and usually more expensive) vendors.

At $work we have to use physical tokens to comply with the security standards of some clients. I forsee such increasingly strict requirements will trickle down to what is deemed critical software infrastructure at some point.

Making "compliance levels" a per-package/dependency-tree instead of per-registry thing is a rather simple change that prevents an increasing security ratchet raising the barrier of entry for hobbyists and less affluent developers.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and locking accounts for which the domain is abandoned.
```
the8472, line 50: This first requires email addresses and then locks people out of their account for losing them.
Some people just want to publish software pseudonymously and use disposable email addresses on sites that force email verification. Normally username+password are sufficient for their purposes and the email is more of a burden.

Instead of making the loss of an email address even worse by also locking people out of an account it's better to only invalidate the email address and let the username+password(+optional additional factors) keep working for as many years as they can maintain their password managers.

And as with my previous comment, instead of burdening everyone it would be more gentle to indicate the level of individual projects without ratcheting up the barrier of entry for everyone using the repository.

docs/principles-for-package-repository-security.md
steiza: Yes, the intent here is that MFA could either be hardware-based or TOTP. What we've seen first-hand at [GitHub's MFA rollout](https://github.blog/2023-03-09-raising-the-bar-for-software-security-github-2fa-begins-march-13/) is that this has substantially reduced account takeover while also providing global access.

docs/principles-for-package-repository-security.md
steiza: On both GitHub and npm, being able to recover access to your account with an email address is a **very** popular feature. I understand your preference, but balancing conflicting requests is one of the challenges of operating systems at scale.

steiza: Do note that this document, as with everything this working group does, is a recommendation, not a requirement. From our README:

> The working group is not a governing body and does not create binding obligations on members

If a package repository decides that a different path is better, that is entirely their choice!

the8472: Public recommendations but reputable organizations have the tendency of being turned into non-negotiable requirements by checkbox-tickers. Therefore careful wording is crucial.

I'm specifically having an issue with the "locking account" wording in this section. Instead if should say something like "invalidate that particular mail address". Some repos have multiple recovery options, support multiple emails or something like that. And having an non-recoverable account into one can at least still log in is still preferable over getting it locked.

steiza: That's fair! I have updated the wording to reflect that what we really want is to remove the ability to recover the account from an abandoned email domain.

docs/principles-for-package-repository-security.md
the8472: The intent is not stated here and I am concerned about this wording contributing to increasingly strict future developments.

You're using the word **requires** here. Today it's perhaps only software TOTP. Tomorrow it'll be a select set of expensive, proprietary hardware tokens. In a decade you can only get those tokens by registering with your national ID, perhaps indirectly by them being tied to a TPM in your phone.
We have seen this with telephony and SIMs. With phone booths it used to be anonymous. Then anyone could buy a SIM with cash, that got you a persistent globally trackable ID that was not directly tied to your name. Now in many countries you can only buy SIMs with an ID. All in the name of fighting crime, terrorism, money laundering or whatever.

With this recommendation and ongoing "critical software infrastructure" efforts in the EU I see this as a small step in that direction.

A step that is easy to mitigate by not making MFA a platform-wide requirement and instead something that individual projects or organizations can opt into. If a project wants to cater to high-assurance clients then they can require MFA from all their maintainers (similar to signed git commits for example) and proudly display their security posture so that their clients can trust the project.

Why would $GOV care that $HOST requires MFA for _all_ users when $GOV only wants to use project $X from $HOST? This requirement just is overly broad. It impacts those who do not share the same values as $GOV.

the8472: Basically, the most-strict needs of some high-assurance projects shouldn't drive the baseline policy of an entire platform.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Builds on behalf of users: For example, Homebrew
+- Only host source code: For example: {index, proxy, sum}.golang.org
+
+## Security Capabilities of Package Repositories
+
+We define 4 levels of security maturity:
+- Level 0 is defined as having very little security maturity.
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
+- Level 2 is defined as having moderate security.
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
```
carols10cents, line 37: What does "smaller" mean in this context, both in terms of what is being measured and what threshold of that measurement is considered to be smaller? That is, does this mean:

- Number of users of the registry?
- Number of libraries hosted in the registry? 
- Number of downloads/people depending on the libraries hosted in the registry?
- Number of maintainers of the registry?

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
```
carols10cents, line 41: Currently, crates.io only allows logging in via OAuth with GitHub-- does this count as having user accounts or not? Some of these items don't apply in this case-- crates.io doesn't have passwords, for example.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and removing the ability to recover an account via an email domain that has been abandoned.
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn
+  - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using leaked credentials. If a user has a leaked credential, the package repository prompts them to change it upon login.
```
carols10cents, line 53: I think this item could use a bit more detail, perhaps in a different, more detailed document-- if credentials have been leaked and the registry knows the user is using leaked credentials, changing the password upon login needs to, say, go through a "password reset" email, correct? Otherwise, anyone who got the leaked credentials could log in with them and change the password themselves? I suppose this goes with the item under level 1 notifying maintainers via email on password changes...

docs/principles-for-package-repository-security.md
the8472:
```suggestion
  - The package repository allows individual packages to require MFA for all maintainers.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
```
carols10cents, line 47: This item seems to be implying that there's also action a user can take if their account has been compromised, and I think it's worth making that explicit here-- that is, adding an item like "The package repository has a support team a user can notify if they suspect their account has been compromised. The support team has a SLA of X for locking and investigating the account".

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 2 is defined as having moderate security.
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
```
carols10cents, line 45: I would add here "and their account recovery procedures that attempts to reduce the likelihood of the package registry support team being susceptible to social engineering attacks" or something along those lines.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+### General Capabilities
+
+Applies to: all package repositories
+
+- Level 1
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities to the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
+  - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
+  - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
```
carols10cents, line 81: I would add here "and documents the support team procedures and SLA for addressing reports of suspicious or malicious packages"

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
+  - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
+  - Malicious packages are published using OSV (e.g. [here](https://github.com/ossf/malicious-packages))
+
+### CLI Tooling
+
+Applies to: all package management ecosystems, although here we are primarily focused on the CLI tooling instead of the registry itself.
+
+- Level 1
+  - The CLI allows installing dependencies that are pinned based on hash, version, etc.
```
carols10cents, line 93: I would explicitly add "source registry" or "location" or similar here, to prevent repeats of issues such as https://nvd.nist.gov/vuln/detail/CVE-2020-36327

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
+  - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
+  - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
+- Level 3
+  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
+  - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
+  - Malicious packages are published using OSV (e.g. [here](https://github.com/ossf/malicious-packages))
+
+### CLI Tooling
```
carols10cents, line 88: Idea for another item under "CLI Tooling": guidance around how the CLI stores/uses/accesses API tokens?

docs/principles-for-package-repository-security.md
the8472: > Publishing the status is something that could be an attack vector itself

Can you explain further?

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 1
+  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
+- Level 2
+  - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
+- Level 3
+  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
+  - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
+  - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
+
+### General Capabilities
```
carols10cents, line 72: Another idea that I think would go under "General Capabilities": a way to view the published artifact on the web without needing to download the package? And more advanced would be a way to view the differences between versions of the artifact?

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
+- Level 2 is defined as having moderate security.
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
```
the8472, line 44:
```suggestion
  - If the user opts to provide an email address the package repository requires users to verify their email address. This ensures that the address will be usable to recover account access.
```

This decouples verification from the decision whether providing an email address should be a (non-)requirement.

If an email is provided it certainly makes sense to verify it. And for many people it's desirable, so it can also make sense to recommend that users provide a valid email address, but in the end there are valid reasons why one might not want to use one. And a user could always opt to sign up with a throwaway email anyway, so enforcing it at signup is not going to really ensure things anyway.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+
+By package repository (sometimes also called package registry) we're referring to the service that stores and index packages, that users and clients can download packages from. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
+
+The taxonomy is because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
+
+The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)) or write specific grant proposals that reference this guidance.
+
+## Taxonomy of Package Repositories
```
carols10cents, line 16: Another sort of package repository is one that doesn't have user accounts and doesn't allow uploads to it directly, but only mirrors another package repository. I think it would be good to have guidance around ensuring the mirror contains the same data as the source, and perhaps guidance around how the mirror removes content that the source has removed because it's malicious?

carols10cents (commented):
Hi! 👋🏻 I'm on the crates.io team and heard from @Turbo87 that you were looking for feedback, so here you are, some feedback based on my experience as a maintainer and "support team" member for crates.io! Happy to provide more clarification on anything :)

Thank you for working on this!

docs/principles-for-package-repository-security.md
ljharb: If i know which packages have MFA and which don’t, i know which ones to attack - the ones that don’t.

the8472: That does not make it an attack vector.

ljharb: And yet, github and npm’s security teams believe it does, which is why they haven’t and likely wont reveal this info on a per-package basis.

the8472: Do you have a source to back that up? Specifically that they consider any policy other than everyone-must-use-MFA an _attack vector_?

And at NPM mandatory MFA appears to be [opt-in at an organization level](https://docs.npmjs.com/requiring-two-factor-authentication-in-your-organization), exactly what I am suggesting.

ljharb: Both services have already announced they’re migrating to 100% mandatory MFA.

the8472: Which still does not substantiate your claim.

ljharb: https://github.com/npm/feedback/discussions/588#discussioncomment-2932982

the8472: That's only talking about "risk to attack", which is not the same as "attack vector". I asked you repeatedly to substantiate your claim that this constitutes an attack vector. Information usable to triaging promising targets does not constitute an attack vector.

Anyway, I consider their risk-based argument lacking nuance/detailed consideration.
For one if the non-mfa population is somewhat large then this information does not steer an attacker towards any particular target. Analogously: Just because a bank in a city block has a safe it does not put any particular private home in the environment at risk of becoming a preferred target for burglars.
It also doesn't consider second-order effects such as pushing people towards using git-based dependencies (which are an option or even the norm in some package managers) if publishing procedures become too onerous for casual users and whether that would have a net-negative effect on security.
There could also be other approaches to mitigating the concern such as only publishing that information if several additional, optional security aspects are also fulfilled. This would dilute the accuracy of the signal.
Yet another option is simply ignoring this concern because the risk is miniscule. Knowing that some niche repos don't use MFA might not be all that useful for an attacker that's looking to compromise a high-value repo.

ljharb: I’m not interested in this argument devolving into pedantry; the fact remains that it is simply not viable to require package or org specific MFA, github and npm will be requiring it wholesale imminently, and this requirement is almost certain to remain as-is in this document.

the8472: > I’m not interested in this argument devolving into pedantry

The difference between an attack vector (which is something that needs to be fixed) and disclosure about security practices (which always is a tradeoff, potentially a tiny one) is significant in my opinion, not pedantry.

I had to ask 3 times for you to even reply to this aspect only to dismiss it as pedantry. I would expect more precise discourse when it comes to designing security documentation for global software infrastructure.

> the fact remains that it is simply not viable to require package or org specific MFA

Do you mean the requirement or publishing that information? The requirement itself is clearly _viable_ and has been for years because that's what NPM is doing right now without the world going up in fire.

There might be reasons to change the policy, but they are likely subtle/marginal ones, that are not about fundamental viability of a policy but about tradeoffs between different viable policies.
Just because github chooses a particular approach does not mean it's the only possible approach.

> and this requirement is almost certain to remain as-is in this document.

Well there was a request for feedback, which is why I'm commenting.

Anyway, I will again reiterate my suggestion that package registries should list the "compliance level" of individual packages (or even individual artifacts) instead of chosing "levels" as a baseline for the entire registry.
By bundling several aspects of those levels into a compliance level it could alsoo discard bits of information an attacker can use. Instead of learning that the package maintainers use MFA specifically they will only learn that the maintainers do a whole bunch of things to reach said compliance level.

So here's a rephrased proposal:

```suggestion
  - The package repository requires MFA for all maintainers.
  - The package repository publishes "security profiles" about individual packages
  - The package repository allows individual packages to require MFA for all maintainers. Enabling it is one of the perquisites of marking an individual package as "level 3" in its security profile.
```

docs/principles-for-package-repository-security.md
di: IMO, it's in terms of the amount of resources available to the registry to add/build these security capabilities.

docs/principles-for-package-repository-security.md
di, line 44: I would add: the repository restricts or otherwise disallows using temporary email addresses, the same email address for multiple accounts, etc.

the8472: I'm opposed to that suggestion. Sometimes I intentionally use temporary email addresses to make a pseudonymous account. Major email providers want to tie your email account to your phone number and such. To avoid that association I use a throwaway to confirm the email validation and then only interact with username+password.

Not being able to recover the account is a choice.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and removing the ability to recover an account via an email domain that has been abandoned.
```
di, line 50: I would argue that the amount of work/effort to do with this probably moves this to Level 3, at least. It's not immediately obvious via a WHOIS whether a domain has been abandoned, or the rightful owner has just changed registrars or their registration information. I think the "right" way to do this is probably requiring something like ACME for all personal domains, but drawing the line here is challenging as well.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn
+  - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using leaked credentials. If a user has a leaked credential, the package repository prompts them to change it upon login.
+- Level 3
+  - The package repository supports passwordless authentication, such as passkeys.
+  - The package repository requires MFA for all maintainers.
```
di, line 56: +1, but also "has a support team to handle account recovery based on a defined process within [some SLO]"

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+### Authorization
+
+Applies to: package repositories that have user accounts and accept built packages.
+
+- Level 1
+  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
+- Level 2
+  - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
+- Level 3
+  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
```
di, line 68:
```suggestion
  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository supports passwordless authentication, such as passkeys.
+  - The package repository requires MFA for all maintainers.
+  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g. top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
+
+### Authorization
+
+Applies to: package repositories that have user accounts and accept built packages.
+
+- Level 1
+  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
```
di, line 64:
```suggestion
  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
  - API tokens are prefixed with a repository-specific identifier
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 2
+  - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
+- Level 3
+  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
+  - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
```
di, line 69:
```suggestion
  - The package repository provides a secret-reporting endpoint for common third-party secret scanning programs like GitGuardian, GitHub, etc. to report leaked credentials matching the known prefix or pattern, which automatically revokes valid credentials.
```

docs/principles-for-package-repository-security.md
di: In my experience, the added burden of spam/malware that comes from allowing temporary email addresses vastly outweighs the benefit of pseudonymous publishing and has a direct effect on the overall security of the repository. A repository team that is overwhelmed with dealing with spam/malware is one that is unable to handle the needs of the legitimate user base, remove spam/malware, or security features in a timely fashion. It's a tradeoff, and the one that improves the security of the repository (which is the goal of this document) is to prohibit temporary email addresses.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
+
+### General Capabilities
+
+Applies to: all package repositories
+
+- Level 1
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities to the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
```
di, line 77:
```suggestion
  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities about the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
```

Assuming this is what's meant by this, not that vulnerabilities would be reported about the software hosted on the repository.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
+  - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
+  - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
+- Level 3
+  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
```
di, line 84: How is "cloud" related here? Should this be "Undergo periodic third-party security audit"?

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
```
di, line 85: Is there maybe a better link that describes what this is and how to use it? A JSON blob is not super helpful.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - Malicious packages are published using OSV (e.g. [here](https://github.com/ossf/malicious-packages))
```
di, line 86:
```suggestion
  - Advisories for malicious packages are published using the OSV schema and aggregated by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages))
```

docs/principles-for-package-repository-security.md
di, line 88: This seems overly broad, are we really just considering specific classes of CLI tooling, e.g. "installers", "builders", "publishers", etc?

docs/principles-for-package-repository-security.md
di: Agreed, something like "pinned based on build/source provenance", but that would probably be Level 2, if not Level 3.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+
+### CLI Tooling
+
+Applies to: all package management ecosystems, although here we are primarily focused on the CLI tooling instead of the registry itself.
```
di, line 90: What should tooling do when it encounters known malware? https://github.com/pypa/pip/issues/5777

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+
+- Level 1
+  - The CLI allows installing dependencies that are pinned based on hash, version, etc.
+- Level 2
+  - The package repository warns of known security vulnerabilities in dependencies, in both the package repository UI and CLI tools when installing.
```
di, line 95: Seems like this should be split between what the repository should do (and moved up into that section) and what the installer should do.

Should this include the ability to automatically remediate known vulnerabilities by upgrading them?

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 3
+  - The CLI has functionality to produce SBOMs.
```
di, line 97:
```suggestion
  - The CLI has functionality to produce SBOMs.
```

I think this could be expanded: produce SBOMs for what? (applications? libraries?) based on what? (a lock file? a source tree?)

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln))
```
di, line 98: Again, this seems like a mix of repository responsibilities (publish sufficiently detailed advisories that make it possible to determine usage) and CLI responsibilities (determine if a given source tree is actually using a vulnerable code path in a dependency with a known vulnerability).

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
+# Principles for Package Repository Security
+
+Authors: [Jack Cable](https://github.com/cablej), [Zach Steindler](https://github.com/steiza)
+Last updated: Dec 2023
+
+## Background
+
+The Securing Software Repositories Working Group (WG) of the OpenSSF is putting together a taxonomy of package repositories and a set of principles for security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
```
di, line 8:
```suggestion
The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
```
Seems weird that this would be in future tense, assuming this describes this document, correct?

docs/principles-for-package-repository-security.md
the8472: What do you consider temporary? A newly created account with a randomized address on a freemail provider that doesn't do phone verification is a temporary address for my purposes.

di: Specifically providers that a) generate email addresses on-demand b) don't require authentication (i.e. an "open inbox"). There is a long list that is maintained here based on that criteria: https://github.com/disposable-email-domains/disposable-email-domains

the8472: I see. I have used both domains in the allow list and in the block list on that repo.
I think it would make sense to spell out the distinction. Broader definitions of "temporary email address" might include anything that's not tied to a person's identity.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+... accept built packages, do builds on behalf of users, or only host source code?
+- Accept built packages: then you have to provide a way to link packages back to source code and build instructions. For example: npm
+- Builds on behalf of users: For example, Homebrew
+- Only host source code: For example: {index, proxy, sum}.golang.org
+
+## Security Capabilities of Package Repositories
+
+We define 4 levels of security maturity:
+- Level 0 is defined as having very little security maturity.
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
```
woodruffw, line 35: Nitpick, makes this a little easier to read (IMO):

```suggestion
- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication, or allowing security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports multi-factor authentication (MFA) via (at minimum) TOTP.
```
woodruffw, line 46: This may be moot given that no major packaging ecosystem uses these (AFAIK), but it might also be worth a negative note here about weak MFA mechanisms like SMS and voice codes.

docs/principles-for-package-repository-security.md
woodruffw: Agreed -- WHOIS is unfortunately a pretty unreliable source of ground truth, and has diminishing value in the world of registrar privacy. 

For better or worse I think the current "state of the art" for email abandonment is bounce detection on delivery, which comes with its own problems. So Level 3 seems appropriate 🙂

docs/principles-for-package-repository-security.md
woodruffw: Big +1 to this -- for rationale, well-known token prefixes help improve the accuracy and performance of techniques like secret scanning (e.g. on GitHub, GitLab, etc.).

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,105 @@
...
+- Level 3
+  - The package repository provisions short-lived identity tokens via  OpenID Connect (OIDC) to prevent needing to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
+  - The package repository API tokens are integrated into common third-party secret scanning programs like GitGuardian, GitHub, etc. This typically requires using a known prefix or pattern to reduce false-positives.
+  - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
+
+### General Capabilities
+
+Applies to: all package repositories
+
+- Level 1
```
woodruffw, line 76: IMO another good "Level 1" item here would be:

```
- The package repository provides per-user and per-project audit logs, allowing users and project maintainers to review recent sensitive actions.
```

(This is what PyPI does, and it's arguably a precursor step to the "event transparency log" noted under "Level 3".)

docs/principles-for-package-repository-security.md
woodruffw: Agreed -- if the doc is meant to be taxonomical, it might make sense to clarify that tooling here is meant to be the "package installer" (based on context).

docs/principles-for-package-repository-security.md
woodruffw: To be more specific: package repositories **should not** use weak MFA methods like SMS codes; the presence of a weak method means that the entire MFA scheme is only as strong as that method (which arguably puts it at "Level 0" in terms of MFA).

docs/principles-for-package-repository-security.md
cablej: How about "and allow"?
```suggestion
- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication and allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
```

docs/principles-for-package-repository-security.md
cablej: How about this?
```suggestion
  - The package repository supports strong multi-factor authentication (MFA) via (at minimum) TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
```

docs/principles-for-package-repository-security.md
cablej: How about this?
```suggestion
  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using known leaked credentials upon login. If a user has a leaked credential, detected upon login, the package repository prompts them to change it. The package manager checks passwords at registration and prevents users from registering with known leaked credentials.
```

docs/principles-for-package-repository-security.md
cablej: Do we have a sense of how many package repositories do this today / whether this would be feasible for smaller package managers?

Either way, I think this is saying 2 separate things, so maybe we say something like this:
```suggestion
  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false-positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
```

docs/principles-for-package-repository-security.md
cablej:
```suggestion
  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
```

docs/principles-for-package-repository-security.md
woodruffw: LGTM!

docs/principles-for-package-repository-security.md
MylesBorins: +1 to level 3 for this... it's actually quite tricky to do this properly

docs/principles-for-package-repository-security.md
MylesBorins: I'd say that consistent internal process for account recovery that is robust to social engineering should be Level 2. Possibly also something about having enough forms of identity verification to enable proper recovery.

If there isn't robust recovery the robust MFA is useless

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
+  - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
```
MylesBorins, line 81: This seems like level 1 tbh. Immutability should be a core requirement for actively maintained packages.

I think there is flexibility here for low used packages (as per the npm policy)

Tombstoning for deletion isn't really neccessary for those packages with low usage imho, those with high usage should be able to be deleted

docs/principles-for-package-repository-security.md
david-a-wheeler: I sometimes differentiate between "registry" and "repository". A "repository" actually *stores* the package contents. A "registry" registers a set of packages, but depends on something else to actually store the data. Would it be useful to make this distinction?

steiza: We settled on "package repositories" in this doc, with the clarification above:

> By package repository (sometimes also called package registry) ...

We last talked about this naming distinction back in November (!) where we looked at [how different ecosystems referred to themselves](https://docs.google.com/document/d/1VuuTnH6Hvsu_kFa_jImPaZe3-xgrn7t3KNeXBpvo5oI/edit?disco=AAABAuzRpXs), and the conclusion we came to is that there isn't consistent terminology.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
+# Principles for Package Repository Security
+
+Authors: [Jack Cable (CISA)](https://github.com/cablej), [Zach Steindler](https://github.com/steiza)
+Last updated: Jan 2024
+
+## Background
+
+The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to be an opinionated document, offering a set of best practices that package repositories should strive to adhere to.
```
cablej, line 8:
```suggestion
The Securing Software Repositories Working Group (WG) of the OpenSSF has identified a taxonomy of package repositories and a set of principles for their security capabilities. This is intended to offer a set of best practices that package repositories should strive to adhere to.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+By package repository (sometimes also called package registry) we're referring to the service that stores and index packages, that users and clients can download packages from. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
```
cablej, line 10:
```suggestion
By package repository (sometimes also called package registry) we're referring to the service that stores and indexes packages, enabling users and clients to download packages. Often users and clients use a CLI tool to download packages, and some ecosystems have multiple popular CLI tools in use. Most of these security capabilities focus on the package repository, but there's also a section on security capabilities the CLI tooling can implement.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+The taxonomy is because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
```
cablej, line 12:
```suggestion
We include the below taxonomy because not all security advice applies to all package repositories. For example, https://proxy.golang.org/ does not manage user accounts, and so has no need of documented account recovery.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)) or write specific grant proposals that reference this guidance.
```
cablej, line 14:
```suggestion
The roadmap of security capabilities can then be used by package repositories to assess gaps, put together fundable improvement lists ([like Python Packaging WG](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md)), or write specific grant proposals that reference this guidance.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+... accept built packages, do builds on behalf of users, or only host source code?
+- Accept built packages: then you have to provide a way to link packages back to source code and build instructions. For example: npm
+- Builds on behalf of users: For example, Homebrew
+- Only host source code: For example: {index, proxy, sum}.golang.org
+
+## Security Capabilities of Package Repositories
+
+We define 4 levels of security maturity:
+- Level 0 is defined as having very little security maturity.
+- Level 1 is defined as having basic security maturity. Package repositories support basic security features like multi-factor authentication and allow security researchers to report vulnerabilities. All package management ecosystems should be working towards this level.
```
cablej, line 35:
```suggestion
- Level 1 is defined as having basic security maturity, which includes supporting basic security features like multi-factor authentication (MFA) and allowing security researchers to report vulnerabilities. All package management ecosystems should be working towards at least this level.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+- Level 2 is defined as having moderate security.
```
cablej, line 36:
```suggestion
- Level 2 is defined as having moderate security, which includes actions like requiring MFA for critical packages and warning users of known security vulnerabilities.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+- Level 3 is defined as having advanced security. These items are more aspirational, especially for smaller package management ecosystems.
```
cablej, line 37:
```suggestion
- Level 3 is defined as having advanced security, which includes actions like requiring MFA for all maintainers and supporting build provenance for packages. This level is more aspirational, especially for smaller package management ecosystems.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Authentication
+
+Applies to: package repositories that have user accounts.
+
+- Level 1
+  - The package repository requires users to verify their email address. This ensures that if a user loses access to their account, they are able to recover it.
+  - The package repository documents their account recovery policy.
+  - The package repository supports strong multi-factor authentication (MFA) via (at minimum) TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
```
cablej, line 46:
```suggestion
  - The package repository supports strong multi-factor authentication (MFA) via, at minimum, TOTP. Solely offering weaker forms of MFA such as SMS, email, or phone call-based MFA would not meet this requirement.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication. This helps users detect if their account has been compromised.
+  - The package repository implements account security measures like brute force prevention (even with 2FA attempts)
+- Level 2
+  - To prevent domain resurrection for account takeover via the recovery process, the package repository detects abandoned email domains. This may look like doing a WHOIS lookup on all registered email domains, and removing the ability to recover an account via an email domain that has been abandoned.
+  - The package repository supports MFA via phishing-resistant means such as WebAuthn
```
cablej, line 51:
```suggestion
  - The package repository supports MFA via phishing-resistant means such as WebAuthn.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository requires MFA for packages deemed critical (e.g. top 1% of packages by downloads).
+  - The package repository integrates with public leaked credential databases such as Have I Been Pwned to detect when users are using known leaked credentials upon login. If a user has a leaked credential, detected upon login, the package repository prompts them to change it. The package manager checks passwords at registration and prevents users from registering with known leaked credentials.
+- Level 3
+  - The package repository supports passwordless authentication, such as passkeys.
+  - The package repository requires MFA for all maintainers.
+  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g. top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
```
cablej, line 57:
```suggestion
  - The package repository requires phishing-resistant MFA for packages deemed critical (e.g., the top 1% of packages by downloads). If possible, the package repository supplies security keys to these maintainers.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Authorization
+
+Applies to: package repositories that have user accounts and accept built packages.
+
+- Level 1
+  - The package repository allows maintainers to provision API keys scoped to specific packages. This allows maintainers to maintain packages via automated workflows without needing to provide their account password.
+  - API tokens are prefixed with a repository-specific identifier
```
cablej, line 65:
```suggestion
  - API tokens are prefixed with a repository-specific identifier.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+- Level 2
+  - The package repository supports role-based access control (RBAC) for maintainers, allowing separate roles for managing users and publishing packages.
+- Level 3
+  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/)
```
cablej, line 69:
```suggestion
  - The package repository provisions short-lived API tokens via an OpenID Connect (OIDC) token exchange to prevent the need to provision long-lived API keys. Example: [trusted publishers](https://docs.pypi.org/trusted-publishers/).
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false-positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
```
cablej, line 70:
```suggestion
  - The package repository API tokens are integrated into common third-party secret scanning programs. This typically requires using a known prefix or pattern to reduce false positives. Furthermore, the package repository provides a secret-reporting endpoint for these secret scanning programs to report leaked credentials, which automatically revokes valid credentials.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository supports providing [build provenance](https://repos.openssf.org/build-provenance-for-all-package-registries) for packages.
+
+### General Capabilities
+
+Applies to: all package repositories
+
+- Level 1
+  - The package repository has published a vulnerability disclosure policy allowing security researchers to report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so. The package repository should strive to remediate reported vulnerabilities within 90 days.
```
cablej, line 78:
```suggestion
  - The package repository has published a vulnerability disclosure policy allowing security researchers to identify and report vulnerabilities affecting the package repository, and receive legal safe harbor when doing so within the requirements of the policy. The package repository should strive to remediate reported vulnerabilities within 90 days.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository takes steps to prevent typosquatting attacks around package names. This may include namespacing, detecting similar names, or other actions.
+- Level 2
+  - The package repository has an unpublish policy to prevent specific versions of a package from being replaced, and, if a package is deleted, prevent others from reusing the same package name and version.
+  - The package repository allows users to report a package as suspicious or malicious via the registry UI, tooling CLI, and via API.
+  - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
```
cablej, line 83: Separated out what was previously in the CLI section.
```suggestion
  - The package repository takes steps to detect malware, including scanning for known malware code snippets and hashes.
  - The package repository warns of known security vulnerabilities in dependencies in the package repository UI.
```

docs/principles-for-package-repository-security.md
cablej:
```suggestion
  - Undergo periodic security reviews of package repository infrastructure (and optionally publish remediations once they are complete)
```

docs/principles-for-package-repository-security.md
cablej: The closest I could find is [https://github.com/npm/registry/blob/master/docs/REPLICATE-API.md](https://github.com/npm/registry/blob/master/docs/REPLICATE-API.md) which appears to be older, not sure if there's anything more descriptive.

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+- Level 3
+  - Undergo periodic cloud security review (and optionally publish remediations once they are complete)
+  - The package repository publishes an event transparency log to enable detection of anomalous behaviors (e.g. [replicate.npmjs.com](https://replicate.npmjs.com/)).
+  - Advisories for malicious packages are published using the OSV schema and aggregated by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages))
```
cablej, line 87:
```suggestion
  - Advisories for malicious packages are published using a standardized machine readable format, such as using the OSV schema and aggregation provided by the OSV service (e.g. [here](https://github.com/ossf/malicious-packages)).
```

docs/principles-for-package-repository-security.md
cablej: Separating out per suggestion:
```suggestion
  - The CLI warns of known security vulnerabilities in dependencies when installing packages.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+### CLI Tooling
+
+Applies to: all package management ecosystems, although here we are primarily focused on the CLI tooling instead of the registry itself.
+
+- Level 1
+  - The CLI allows installing dependencies that are pinned based on hash, version, etc.
+- Level 2
+  - The package repository warns of known security vulnerabilities in dependencies, in both the package repository UI and CLI tools when installing.
+- Level 3
+  - The CLI has functionality to produce SBOMs.
```
cablej, line 98:
```suggestion
  - The CLI has functionality to produce SBOMs.
  - The CLI has functionality to, where possible, automatically remediate known vulnerabilities in dependencies by upgrading them.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln))
```
cablej, line 99:
```suggestion
  - The package repository is able to reduce false positives of identified vulnerabilities using static analysis to understand whether a vulnerable code path is actually reachable. (e.g. [Golang's vulncheck feature](https://go.dev/security/vuln)).
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Resources / Inspiration
+
+- <https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md>
+- <https://github.com/ossf/wg-securing-software-repos/blob/main/survey/2022/results.csv>
+- <https://docs.npmjs.com/threats-and-mitigations>
+- <https://www.microsoft.com/en-us/research/uploads/prod/2018/03/build-systems.pdf> (as a framework for comparing options in a problem space)
```
cablej, line 106:
```suggestion
- <https://www.microsoft.com/en-us/research/uploads/prod/2018/03/build-systems.pdf> (as a framework for comparing options in a problem space)

Note: CISA does not endorse any commercial entity, product, company, or service, including any entities, products, or services linked within this document. Any reference to specific commercial entities, products, processes, or services by service mark, trademark, manufacturer, or otherwise, does not constitute or imply endorsement, recommendation, or favoring by CISA.
```

docs/principles-for-package-repository-security.md
```diff
@@ -0,0 +1,110 @@
...
+- No: For example: {index, proxy, sum}.golang.org
+
+... accept built packages, do builds on behalf of users, or only host source code?
+- Accept built packages: then you have to provide a way to link packages back to source code and build instructions. For example: npm
+- Builds on behalf of users: For example, Homebrew
+- Only host source code: For example: {index, proxy, sum}.golang.org
+
+## Security Capabilities of Package Repositories
+
+We define 4 levels of security maturity:
```
cablej, line 33:
```suggestion
We define 4 levels of security maturity. Similar to [SLSA](https://slsa.dev/spec/v1.0/levels), levels of this framework are split into four tracks enumerated in the below sections: Authentication, Authorization, General Capabilities, and CLI Tooling. All tracks may not apply to all package repositories, as detailed above. Having these tracks allows package repositories to assess their security across various dimensions. In order to be considered to meet a certain level of security maturity, a package repository must meet all actions in all applicable tracks. 
```

steiza:
Thank you everyone for your incredible feedback! The diversity of perspectives from different package repositories has made it clear there isn't a single linear path that makes sense for everyone. And yet I think this makes a map of security capabilities even more valuable as a way to share learnings, and for each ecosystem to consider what parts make sense for them. This document has improved considerably in the past two months, so again, thank you for your feedback.

We're going to cut a v0.1 release of this document today. While it feels great to hit that milestone, there's additional feedback on this pull request that we didn't get to and we need to still follow up on. I don't know exactly what that process will look like, but right now I think the most likely option is we'll open up a v0.2 pull request soon, to address feedback raised here as well as wider public feedback we get from the release of v0.1.

Issue #47, opened 2024-09-04
Create "Binary Transparency for Artifact Registries" guide


Hayden-IO:
Binary Transparency (BT) adds cryptographic immutability and discoverability to package identifiers and metadata, and when applied to artifact registries, prevents registries from acting maliciously, e.g. by tampering with existing artifact versions or serving new malicious versions.

As usage of Binary Transparency is not widespread and overlaps with other related supply chain security projects, this guide will serve as an overview to BT, discussing threats mitigated by BT, case studies of where BT has been applied, and how BT compares to other solutions like code signing or attestations. If you attended the WG meeting where I [presented on BT](https://docs.google.com/presentation/d/15JF-qi3wxgtU2m3UQLbIT7KPEYLtHcVftyp1LlN5P6I/edit?resourcekey=0-fjZCHpnmCVHW-LLPT8QISw#slide=id.p), this guide will be a far more detailed analysis of BT and should serve as a the source of truth for how we want to discuss BT applied to artifact registries.

An implementation of this guide will follow shortly, with a goal of turning the API into a spec for [C2SP](https://github.com/C2SP/C2SP).

PR #48, opened 2024-09-04
Create Binary Transparency for Artifact Registries guide

haydentherapper:
Fixes https://github.com/ossf/wg-securing-software-repos/issues/47

### DIFF
--- a/docs/binary-transparency-for-artifact-registries.md
+++ b/docs/binary-transparency-for-artifact-registries.md
@@ -1,0 +1,160 @@
+# Binary Transparency for Artifact Registries
+
+Authors: [Hayden Blauzvern (Google)](https://github.com/haydentherapper)
+
+Last updated: September 2024
+
+This document defines Binary Transparency for artifacts registries to mitigate the risk of registry compromise by external attackers or insiders.
+
+## What is Binary Transparency?
+
+[Binary Transparency](https://binary.transparency.dev/) (BT) is a security solution that adds cryptographic immutability and discoverability to package identifiers and metadata, which in turn prevents tampering of existing versions and supports detection of unauthorized new versions.
+
+BT is backed by a cryptographic data structure that is immutable and append-only. When publicly accessible, this structure becomes a transparency log, providing transparency into every committed decision present in the log. Each entry into the log receives an inclusion proof, a cryptographically verifiable confirmation of the inclusion of the entry, which is checked during artifact verification.
+
+In the context of artifact registries, BT provides an assurance that all users always receive the same binary for a given package identifier, because users require the registry to transparently commit to all binaries it serves. If a registry is compromised or compelled to serve a different binary, then it must do so for all of its users and for all time, and because this decision must be recorded in a public transparency log, the malicious action can be audited and detected.
+
+## Applying Binary Transparency to Artifact Registries
+
+BT applies to any package distribution method using logically immutable package identifiers[^1]. This includes most OSS package ecosystems, such as Go, PyPI (Python), Cargo (Rust), npm (JavaScript), Maven Central (Java), and CI/CD registries such as GitHub Releases[^2]. In other words, users identify and fetch a package by `<package_name>`, `<version_id>`, and `<filename>` (or similar), referred to as a package identifier, and receive back some artifact. Logically, this process should *always* and *globally* return the same artifact[^3]. However, most ecosystems lack sufficient cryptographic security controls to ensure that this always occurs, making them susceptible to tampering by compromised registries, mirrors, or distribution channels. This is where BT fits in.
+
+BT makes package identifiers immutable and transparent:
+
+* A package registry will publish a commitment that a specific package identifier maps to a cryptographic digest. This commitment creates an immutable mapping between the identifier and digest. The mapping can only be done per package identifier and can never be updated or deleted.  
+  * At upload time, the producer verifies that the entry was recorded correctly.  
+  * At download time, the consumer verifies that the artifact they received matches the digest found in the log through an offline cryptographic inclusion proof.  
+* Since the BT log is public, the mapping is publicly auditable.  
+  * Artifact producers will verify that the log only contains artifacts they've released.  
+  * Third-party auditors will verify that there exists only one instance of a cryptographic hash per package identifier.
+
+## Threats Mitigated by Binary Transparency
+
+BT mitigates threats associated with registry or mirror compromise or compulsion to serve a malicious artifact. Specifically:
+
+### Tampering of existing versions
+
+The threat is that a client requests a package by its immutable package identifier, but receives an artifact that does not correspond to the artifact the producer originally uploaded. This tampering can occur not only from the registry, but also from a mirror, network-level interference, or from storage.
+
+BT prevents this completely. Once an identifier has been committed to the registry, clients only accept that corresponding digest. A producer can prevent the registry from tampering with the artifact at upload time by requiring proof that the digest has been committed to the BT log correctly, and a consumer must not accept an artifact without verifying the inclusion proof.
+
+### Unauthorized publication of new versions
+
+The threat is that the client asks the registry for the latest version and gets back a version identifier that the producer did not actually publish. For example, if the latest legitimate version of a given package is 1.0.0, a compromised registry might return 1.0.1 with malicious content.
+
+BT does not prevent this threat outright, but makes this type of attack discoverable since the malicious action must be committed to a log. In this example, the malicious artifact could be discovered by a producer that is monitoring the BT log for new instances of their artifact. We recognize that monitoring could require active participation from all artifact maintainers and that there are current gaps in tooling[^4], but we believe that forcing malicious behavior to be publicly auditable acts as a significant deterrent against such behavior and provides non-repudiation for the registry's actions.
+
+In conjunction with a secure update framework such as [The Update Framework](https://theupdateframework.io/) (TUF), artifact updates could be made transparent, immutable, and impossible to rollback or forward without discovery by a producer.
+
+### Covert publication of malicious versions
+
+The threat is that a producer is compelled to produce multiple versions of a given package, one "good" and the other "bad", e.g. two versions of 1.0.0, or a good 1.0.0 release made widely available and a bad 1.0.0-epsilon version which is distributed in a more targeted fashion.
+
+BT makes such attacks discoverable since both versions must be present in the log. 
+
+## Gaps Still Present with BT
+
+### Typosquatting/Dependency Confusion
+
+Typosquatting is when a malicious artifact is uploaded with either a name similar to a popular artifact, and dependency confusion is when an artifact is uploaded to a public registry where it exists only in a private registry with the intent of confusing a dependency resolution system. In either case, the threat is a consumer depending on an incorrect artifact. BT does not solve this problem, as BT cannot detect artifacts with similar names. BT only guarantees that a request for a specific artifact receives a globally consistent response.
+
+Depending on the package identifier, verifiers could use BT log entries to flag suspicious names, as BT would make all package identifiers discoverable.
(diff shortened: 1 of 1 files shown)


docs/binary-transparency-for-artifact-registries.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Applying Binary Transparency to Artifact Registries
+
+BT applies to any package distribution method using logically immutable package identifiers[^1]. This includes most OSS package ecosystems, such as Go, PyPI (Python), Cargo (Rust), npm (JavaScript), Maven Central (Java), and CI/CD registries such as GitHub Releases[^2]. In other words, users identify and fetch a package by `<package_name>`, `<version_id>`, and `<filename>` (or similar), referred to as a package identifier, and receive back some artifact. Logically, this process should *always* and *globally* return the same artifact[^3]. However, most ecosystems lack sufficient cryptographic security controls to ensure that this always occurs, making them susceptible to tampering by compromised registries, mirrors, or distribution channels. This is where BT fits in.
+
+BT makes package identifiers immutable and transparent:
+
+* A package registry will publish a commitment that a specific package identifier maps to a cryptographic digest. This commitment creates an immutable mapping between the identifier and digest. The mapping can only be done per package identifier and can never be updated or deleted.
```
Eh2406, line 23: Most of the ecosystems listed above have some form of package deletion. Many provide a way for a user to request one of the packages be deleted, in a self-service manner, under various safeguards. Even the ones that don't will remove an artifact when compelled by a court.  Generally, if they allow deletion they also allow there to be a new owner who is allowed to publish new packages at the same version number. Unfortunately, these corner cases cannot be ignored. How does the system accommodate this reality?

haydentherapper: Great question! In one of the [footnotes](https://github.com/ossf/wg-securing-software-repos/pull/48/files#diff-8e0542c5388ca2387892b0ed56a55b81130d24fd3a094c59988be53a22021accR158), I touch on if the ecosystem supports yanking a package. Supporting package deletion wouldn't be at odds with binary transparency, _as long as_ the ecosystem wouldn't allow package identifier reuse, e.g. if `package@v1.2.3` is deleted from the registry, then the next package version **must** be `v1.2.4` or higher (or more accurately, a version that's never been used before, BT is not opinionated on ever-increasing versions, that's an ecosystem decision).

Lockfile support in an ecosystem also should require this to be true, otherwise yanking and recreating a package at a given version would break existing lockfiles.

A question for the registries, do you allow version reuse? At a glance:

* [npm does not allow version reuse](https://docs.npmjs.com/policies/unpublish#considerations)
* [PyPI does not allow version reuse if a file was deleted](https://pypi.org/help/#file-name-reuse)
* [Maven Central doesn't support deletion](https://central.sonatype.org/faq/can-i-change-a-component/)
* [RubyGems supports yanking, but unspecified on version reuse](https://guides.rubygems.org/removing-a-published-gem/)
* [NuGet supports unlisting, but not deleting iiuc](https://learn.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages)
* Go catches version reuse through its [proxy](https://sum.golang.org/) since it runs a binary transparency log. For example, if a tag is deleted and recreated pointing to a different commit hash, the Go proxy will throw an error.

woodruffw: To tack onto what @haydentherapper said: PyPI's behavior is subtle in that *filenames* are always unique and immutable on PyPI, but *releases* themselves are not. In other words: a project `foo` that gets deleted or turned over to a new user can't overwrite `foo-1.2.3.tar.gz` if that distribution file was already uploaded by a previous maintainer, but the new maintainer *might* be able to upload `foo-1.2.3-py3-none-any.whl` or similar if no wheel was previously uploaded for that version of `foo`. 

In practice this means that resolving `foo==1.2.3` isn't guaranteed to be stable for a given host, since the new maintainer can always upload a new (unique) file to the old version that's more specific/matches the host's target configuration more precisely.

david-a-wheeler:
[woodruffw](https://github.com/woodruffw) [16 hours ago](https://github.com/ossf/wg-securing-software-repos/pull/48#discussion_r1744482051) said:

> To tack onto what @haydentherapper said: PyPI's behavior is subtle in that filenames are always unique and immutable on PyPI, but releases themselves are not. In other words: a project foo that gets deleted or turned over to a new user can't overwrite foo-1.2.3.tar.gz if that distribution file was already uploaded by a previous maintainer, but the new maintainer might be able to upload foo-1.2.3-py3-none-any.whl or similar if no wheel was previously uploaded for that version of foo.

> In practice this means that resolving foo==1.2.3 isn't guaranteed to be stable for a given host, since the new maintainer can always upload a new (unique) file to the old version that's more specific/matches the host's target configuration more precisely.

That sounds extremely problematic to me. I expect that repositories must sometimes remove packages of a given version (e.g., by court order), but I think most users would expect that a given version# would be stable. This loophole is a great way to hide attacks. Is this functionality critical to PyPI somehow? Could PyPI be changed to prevent it (at least, say, after a day or so of the "initial" version being uploaded)?

woodruffw:
> This loophole is a great way to hide attacks. Is this functionality critical to PyPI somehow? Could PyPI be changed to prevent it (at least, say, after a day or so of the "initial" version being uploaded)?

There's a long-ish thread on the current behavior here: https://discuss.python.org/t/restricting-open-ended-releases-on-pypi/43566

The TL;DR of it is that PyPI having "open-ended" releases is currently relied upon for some packaging workflows, e.g. there are maintainers who update their releases to contain wheels for new versions of Python rather than publishing an entirely new version with no functional changes. There's also some debate about how serious the vector is, given that (1) the attacker can only upload new files, not overwrite existing release files, and (2) could always just make a new release instead, given that Python as an ecosystem tends to avoid exact version-pinning.

But apart from that, +1 to everything @sethmlarson said, especially drafting -- there _is_ a PEP that enables support for drafting on PyPI and other indices, and I've (very) recently been given the resources to begin work on actually implementing it 🙂

di:
> I'm not sure how often this happens in practice and whether or not it's worth the additional risk because I don't maintain many packages that have these requirements.

It happens quite often!

> That sounds extremely problematic to me. I expect that repositories must sometimes remove packages of a given version (e.g., by court order), but I think most users would expect that a given version# would be stable. This loophole is a great way to hide attacks.

The risk is entirely mitigated by using lockfiles or hash-pinned requirements files.

> I think adding support for "draft" releases would make removing this property of an index viable, but even then I am not sure of the impact for maintainers, needs more studying to be sure.

Draft releases definitely helps with the "artifacts come from different places at different times" issue, and is desirable for other reasons, but it doesn't resolve the "build new artifacts for old releases against new Python versions, ABIs or platforms" so it is not a panacea here.

steiza (commented):
Generally this looks good to me!

Let's start out with some easier items:

- Time and time again this working group struggles with terminology. In keeping with the working group name, we usually use "repository" and "software repository" where this document uses "registry" and "artifact / package registry". Let's standardize on one term in the document, and it might be a good idea to have a terminology / definition section near the front.

- In this pull request, can we add a link to this doc on both https://github.com/ossf/wg-securing-software-repos/blob/main/README.md and https://github.com/ossf/wg-securing-software-repos/blob/main/docs/index.md?

- Now for the tricky one! Generally speaking, our other guides cover an existing successful implementation and high-level guidance on how other repositories can implement it. Of course, Go does have binary transparency (as this doc mentions), but the `Design Ideas` section (where I think the implementation guidance would live) is TBD in this version. Do you want to land a first version of this doc and flesh that out further later? How do you want to proceed here?

david-a-wheeler:
> Time and time again this working group struggles with terminology. In keeping with the working group name

It happens often, I'm afraid.

> we usually use "repository" and "software repository" where this document uses "registry" and "artifact / package registry".

Obviously people different in what they mean by terms. For example, I should note that I don't normally use these terms as synonyms. Here's how I normally use the terms:

A (package) *repository* stores information. PyPI and CPAN, for example, actually *stores* the packages that can be installed by a package manager. A source repository stores the source code that you might use (e.g., to build).

A *registry* is an "official" record of "where to get the information" - but often doesn't store the data itself. Registries redirect users to 1+ repositories. Depending on the registry, different components in the registry might be served by different repositories. I know quicklisp works this way, and I think others do too.

I don't claim everyone uses the terms the same way. That's part of the challnege here :-).

> Let's standardize on one term in the document, and it might be a good idea to have a terminology / definition section near the front.

100% agree. Trying to get everyone to change terminology throughout the world to the same thing is er, hard. Documenting definitions of key terms, as they are used in the document, definitely sounds like a way forward.

di:
> Now for the tricky one! Generally speaking, our other guides cover an existing successful implementation and high-level guidance on how other repositories can implement it.

I think we're pretty close to this with [PEP 740](https://peps.python.org/pep-0740/) for PyPI, but I agree that we might want to wait to publish until that has been fully baked and any unforseen issues are sorted out.

docs/binary-transparency-for-artifact-registries.md
```diff
@@ -0,0 +1,160 @@
...
+### Deployment
+
+* Binary Transparency must incorporate all elements of a complete transparency ecosystem integration, with strong offline proofs that bundle an inclusion proof and co-signed witness checkpoints. Witnesses will verify the consistency of the log, that the log remains append-only, as is the requirement for any transparency ecosystem.  
+* A log must restrict publishing to only the registry operator. Otherwise, it is not possible to monitor the log for unexpected artifact hashes and hold the registry accountable. A registry may choose to operate the log itself, but the log could also be operated by another entity, as long as in either case, log publication is restricted to the registry operator.
+
+## Claimant Model
+
+Terminology defined in [Trillian documentation](https://github.com/google/trillian/tree/master/docs/claimantmodel).
+
+* The Claim is "There exists at most one mapping between a given artifact's package identifier and its cryptographic hash".
```
mhutchinson, line 147: This claim doesn't encode the filename.

docs/binary-transparency-for-artifact-registries.md
```diff
@@ -0,0 +1,160 @@
...
+
+* Binary Transparency must incorporate all elements of a complete transparency ecosystem integration, with strong offline proofs that bundle an inclusion proof and co-signed witness checkpoints. Witnesses will verify the consistency of the log, that the log remains append-only, as is the requirement for any transparency ecosystem.  
+* A log must restrict publishing to only the registry operator. Otherwise, it is not possible to monitor the log for unexpected artifact hashes and hold the registry accountable. A registry may choose to operate the log itself, but the log could also be operated by another entity, as long as in either case, log publication is restricted to the registry operator.
+
+## Claimant Model
+
+Terminology defined in [Trillian documentation](https://github.com/google/trillian/tree/master/docs/claimantmodel).
+
+* The Claim is "There exists at most one mapping between a given artifact's package identifier and its cryptographic hash".  
+* The Statement is the package identifier and digest.
```
mhutchinson, line 148: This isn't a definition of a Statement. A statement must define the encoding, and must include all of the components committed to by the Claim.

docs/binary-transparency-for-artifact-registries.md
mhutchinson: Does "package identifier" point at the triple {module, version, filename} ?

mhutchinson: I see that the top of the doc says that this triple is the package identifier. Closing this comment.

docs/binary-transparency-for-artifact-registries.md
mhutchinson: I can't tell from this proposal whether the plan is that all adopters take the same Statement (and thus run identical logs), or whether you're trying to describe an approach, and then allow each artifact registry fit specifics such as the statement to their needs?

PR #51, opened 2025-03-05
Package Yanking Guidance 

hayleycd:
Please see this [TAC Issue](https://github.com/ossf/tac/issues/414). 

This is no longer a draft. I am open to any suggestions or changes.


hayleycd:
@steiza Thanks so much! I will take this feedback and will be pushing a draft in the next few days.

hayleycd:
@steiza the draft is still WIP, but there is actual text instead of just notes. The complete draft will be ready in the next couple of days, but please take a look if you would like to.

di:
@hayleycd Is this ready to be marked as "ready for review"?

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,42 @@
...
+Last updated: March 2025
+
+NOTES: To be deleted when published. 
+
+* Opening
+    * Left pad incident
+    * Package managers need to protect the integrity of their registries. Package managers provide a valuable service of delivering open source packages and unchecked package deletion would render package managers unable to perform the service for which they were designed. 
+
+* Types of Package deletion policies
+    * Allow for no deletion
```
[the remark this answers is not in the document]
david-a-wheeler: It's not just impractical; it's illegal in many jurisdictions. If your package includes child porn, and your repo refuses to delete it after a court order, the authorities may choose to shut down the site & jail those who run it.

I recommend changing this to:

```suggestion
    * Allow for no deletion unless required by law/court order
```

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,42 @@
...
+                * Etc
+        * Pros
+            * People may be experimenting with how to create a package on the registry, but they don't actually want to publish a package (or at least, not yet). Not allowing for deletion can discourage new people from engaging. 
+            * People make mistakes, they may have a typo in the package name, maybe they published the wrong package, etc
+            * Automated--package manager maintainers do not have to manually intervene
+            * Limits potential scope of problem. May break _some_ builds, but nothing on the scale of the left pad incident
+        * Cons
+            * How are these limits determined? Setting them too loose increases the risk of issues, keeping things too tight increases the amount that maintainers will need to intervene. If the settings are too tight the registry is a little cluttered and includes packages that people aren't actually using, etc. 
+    * Allow for version yanking
+    * Deprecation
```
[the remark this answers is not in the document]
david-a-wheeler: It's usually a bad idea to completely yank a vulnerable version. The vulnerability might not be exploitable in context, and many organizations need a lot of time to update components.

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,156 @@
...
+
+In 2016 after a dispute about the naming of an open source package, a maintainer deleted all the packages he had published to npm, including a small package called `left-pad`. Although `left-pad` was only about a dozen lines of code, it had impressive reach. Prior to removal, it had over 15 million downloads and was used by thousands of open source projects--including many that are critical to the npm ecosystem. The deletion of `left-pad` caused build failures for projects large and small and rendered a large portion of the internet unnavigable. Npm made the decision to republish the package.  
+
+After the `left-pad` incident, npm updated their package deletion policy to prevent similar failures and other package managers followed suit. In the over 9 years since the incident, package managers have continued to adjust their package deletion policies to balance the sometimes competing interests of open source maintainers, open source consumers, and the package ecosystem itself. 
+
+This paper discusses the components of modern package deletion policies. It is not meant to be a sample deletion policy, but a discussion of things to consider when crafting such a policy.
```
david-a-wheeler, line 9: Are we talking about deleting all versions of a package? Or only specific versions? That needs to be clear early on. I'm guessing you want to discuss both.

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,156 @@
...
+## Deletion policy considerations
+
+If we consider a hypothetical package deletion policy, it would exist on a spectrum somewhere between never allowing packages to be deleted to allowing packages to be deleted without any restrictions. Both ends of this spectrum are impractical. 
+
+* Never allowing for package deletion could result in the publishing of malicious, abusive, illegal, or copywritten content without any recourse.
+* Allowing for unrestricted package deletion could result in build failures like those seen in the `left-pad` incident, malicious packages snapping up a recently vacated namespace, and other undesireable behavior. 
+
+So when crafting a deletion policy, we want to stay away from the two extremes, but what needs to be considered?
+
+Whether or not a package manager allows for users to delete packages in some cases, it should allow for the package manager maintainers to manually intervene and delete packages. This allows the maintainer team to protect the ecosystem from malicious, abusive, illegal, or copywritten content.
```
david-a-wheeler, line 20:
```suggestion
Whether or not a package manager allows for users to delete packages in some cases, it should allow for the repository maintainers to manually intervene and delete packages. This allows the maintainer team to protect the ecosystem from malicious, abusive, illegal, or copywritten content. 
```

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Time
+
+How long ago was the package published? Although time since publication is not a perfect metric for ecosystem impact (maybe a package has been around for years but only has a handful of users), there are a number of use cases where this metric is very relevant. For example, the user could have:
+
+* Published the wrong package
+* Published the package with an unintended name (or a typo)
+* Published a test package to understand and experience the package publishing process
+
+In these cases, it is very useful for the package manager to have a no hassle deletion policy as long as the package has been published for less than a certain amount of time.
```
david-a-wheeler, line 40:
```suggestion
In these cases, it is very useful for the package manager to have a no hassle deletion policy as long as the package has been published for less than a certain amount of time (e.g., 7 days). 
```

david-a-wheeler: Give an example. Feel free to change the example, but giving an example would help clarify things.

hayleycd:
> @hayleycd Is this ready to be marked as "ready for review"?

It is almost ready. I am going to address the comments that came in today. Remove my notes, and add a list of resources. Then I will move it out of draft.

docs/package-yanking-policies.md
hayleycd: There is a discussion of both, but that is a good point for the opening. I will adjust.

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,149 @@
+# Crafting a package deletion policy
+
+Last updated: April 2025
```
steiza, line 3: Minor: if you'd like this work to be attributed to you, you can add an author line like we have on https://github.com/ossf/wg-securing-software-repos/blob/main/docs/trusted-publishers-for-all-package-repositories.md?plain=1#L3 or https://github.com/ossf/wg-securing-software-repos/blob/main/docs/principles-for-package-repository-security.md?plain=1#L3

steiza (approved):
This is great!

In this pull request, do you also want to add a link under `Implementation Guidance` on https://github.com/ossf/wg-securing-software-repos/blob/main/docs/index.md?plain=1#L10? Then a link to this page will show up on https://repos.openssf.org/.

miketheman (commented):
I enjoyed reading through this, well done! I had some notes/suggestion inline, and also wanted to drop this resource: https://sembr.org/
It might prove useful for longer-form documents that may only change a word or two on a line, and thus prevent having to update an entire paragraph as a result. The output rendering ought not change.

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,149 @@
...
+
+After the `left-pad` incident, npm updated their package deletion policy to prevent similar failures and other package managers followed suit. In the over 9 years since the incident, package managers have continued to adjust their package deletion policies to balance the sometimes competing interests of open source maintainers, open source consumers, and the package ecosystem itself.
+
+This paper discusses the components of modern package deletion policies. It is not meant to be a sample deletion policy, but a discussion of things to consider when crafting such a policy. It also includes a discussion of alternatives to unpublishing a package.
```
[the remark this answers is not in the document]
hayleycd: Good point. I have rephrased the sentence.

docs/package-yanking-policies.md
```diff
@@ -0,0 +1,149 @@
...
+* [Unpublishing Packages from the Registry](https://docs.npmjs.com/unpublishing-packages-from-the-registry)
+* [npm Unpublish Policy](https://docs.npmjs.com/policies/unpublish)
+
+NuGet
+
+* [Deleting Packages](https://learn.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages)
+
+Python
+
+* [What's a "yanked" release?](https://pypi.org/help/#yanked)
```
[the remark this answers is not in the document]
hayleycd: Thanks! I like that change.

hayleycd:
> I enjoyed reading through this, well done! I had some notes/suggestion inline, and also wanted to drop this resource: https://sembr.org/ It might prove useful for longer-form documents that may only change a word or two on a line, and thus prevent having to update an entire paragraph as a result. The output rendering ought not change.

Thanks! I learned something new today. I didn't know that is how markdown rendered things and that's really useful for review/revision.

docs/package-yanking-policies.md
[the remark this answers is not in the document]
hayleycd: I agree. Thanks!

hayleycd:
Thanks @miketheman @david-a-wheeler @di and @steiza! I appreciate your feedback. 

@steiza I have addressed comments and I've added a link to the index page.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+Package deletion is a good solution for some of those reasons (publishing the wrong package, making a typo, experimenting with the package manager), but not necessarily for others.
+
+What if a package (or version series) is no longer in active development?
+What if a specific version has a security issue?
+In these cases, the community should be informed of the issues, and possibly prevented from using a specific version, but deleting the package outright is not a tenable solution.
+So what can package managers do in these cases?
+
+### Yanking
+
+"Yanking" refers to soft deleting a compromised package version.
```
di, line 132: I think "soft deleting" is probably the wrong term to use here. If "deleting" means "the package is completely gone", "soft deleting" would be "the package is completely gone, but can be fully restored as it was prior to deletion".

Yanking is even softer than soft deletion, because in many ways, the package still exists. A better way to describe this may be something like "yanking is a way to deprioritize an installer's preference for a given release or file, relative to other releases or files in a project".

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+The version still exists, but by marking the version as yanked, installers will ignore the version unless it is the only release that matches a version specifier.
+If the yanked version is explicitly required, it will be installed and will not break the build.
+Maintainers should consider yanking a release if it is broken, unstable, or contains a very severe vulnerability.
+Yanking will not completely guarantee that no one is using the compromised version, but it will reduce the prevalence of the corrupted version.
```
di, line 136:
```suggestion
Yanking will not completely guarantee that no one is using the yanked version, but it will reduce the installer preference to install it.
```

di (approved):
Looks great! Just some high level feedback on terminology:

- This uses the terms "package registries" and "package managers" somewhat interchangeably. Generally, I would recommend using "package repositories" or "package index" to refer to an index with a collection of packages (e.g. PyPI) and "package installers" to refer to a tool a user would use to manage their packages locally (e.g. pip).
- We may also want to be more clear when we use the word "maintainers" -- while the team that _runs_ a repository could also be considered maintainers, a better term might be "repository administrators", which avoids ambiguity when referring to end users as maintainers.

docs/package-deletion-policies.md
david-a-wheeler: I agree with @di that the terminology should be clear. I would use the term "package manager" not "package installer" for the local software used to install/uninstall/maintain the software; I think package manager is the more common term.

In some ecosystems the list of managed packages (what I would call the package registry or package index) is separate from where the packages are stored (what I would call package repository). In some other cases one system fulfills both roles, which leads to confusion. I suspect in most cases the concern should focus on the package registry (aka package index) - if it's gone from the *repository* then you can't get it at all.

hayleycd: Ah ok. That will clarify things. I'll take a quick look through the report and update terms.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+
+Whether or not a package manager allows for users to delete packages in some cases, it should allow for the package manager maintainers to manually intervene and delete packages.
+This allows the maintainer team to protect the ecosystem from malicious, abusive, illegal, or copyrighted content.
+
+But what if a user wants to delete a package they published?
+Although it's possible to require such cases to go to the maintainer team to be manually reviewed and removed, that is not a very efficient policy.
+Such a policy would place a burden on the (already likely overworked) maintainer team, and the ecosystem would be cluttered with the unwanted packages until they are manually removed.
+
+Allowing a user to delete their package in certain cases is therefore good for the user, the package manager maintainer team, and the package ecosystem at large.
+But what factors need to be considered to avoid a `left-pad` style incident?
```
joelverhagen, line 37: ... while still avoiding illegal or unsafe content.

I think you have to hold both in each hand, pretty much all the time. For example if we had to left-pad the world because of some DMCA'd or malicious content, I think you'd have to as the repo admin.

NuGet.org has deleted relatively popular content that we found was infringing copyright.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+
+* Published the wrong package
+* Published the package with an unintended name (or a typo)
+* Published a test package to understand and experience the package publishing process
+
+In these cases, it is very useful for the package manager to have a no hassle deletion policy as long as the package has been published for less than a certain amount of time.
+For instance, npm allows maintainers to delete their packages within the first 72 hours after publication.
+After 72 hours additional criteria must be met.
+
+### Downloads
```
joelverhagen, line 59: Might want to make this a little more generic or qualify in the paragraph. Usage statistics vary from ecosystem to ecosystem. For example VS Code uses Installs. In NuGet, we know not all downloads are created equal. Perhaps it's worth saying how you assess the breadth of package usage may vary by ecosystem.

docs/package-deletion-policies.md
joelverhagen: I agree we need to formalized wording. To me, yanking is unclear and we use the term deletion to refer to soft deletion in NuGet (with a very rare hard deletion). For us, soft deletion means it is really deleted, but we still have a record of it and the package name is blocked.

I have found that soft delete and hard delete are commonly used in software engineering such as database design so it's easier to describe that for packages. But as long as we define terminology at the talk of the document, I think we can go with whatever and even provide comparisons on common ecosystems like list of synonyms even.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+What if a package (or version series) is no longer in active development?
+What if a specific version has a security issue?
+In these cases, the community should be informed of the issues, and possibly prevented from using a specific version, but deleting the package outright is not a tenable solution.
+So what can package managers do in these cases?
+
+### Yanking
+
+"Yanking" refers to soft deleting a compromised package version.
+The version still exists, but by marking the version as yanked, installers will ignore the version unless it is the only release that matches a version specifier.
+If the yanked version is explicitly required, it will be installed and will not break the build.
```
joelverhagen, line 134: Oh weird, for this feature we call it unlisting on NuGet and unpublishing on VS Marketplace.

Yanking doesn't sound at all appropriate for malicious or illegal content if it can still be downloaded.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+For example, Go uses the term "retract" and the retract directive can indicate that a package version (or version range) should not be used.
+
+Regardless of the term used, the following sets these policies apart from a deletion policy in the following ways:
+
+1. The yanked version is still available.
+2. The installer will try to avoid using yanked versions.
+3. The package consumer can make an informed decision about whether to use the yanked version.
+4. Yanking a version will not break a build.
+
+### Deprecation
```
joelverhagen, line 150: You could maybe lump end of life status as part of this. In many contexts the two are used interchangeably but on ecosystem to ecosystem. For the purposes of this document, I think they can be lumped together to improve comprehension.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+This allows the maintainer team to protect the ecosystem from malicious, abusive, illegal, or copyrighted content.
+
+But what if a user wants to delete a package they published?
+Although it's possible to require such cases to go to the maintainer team to be manually reviewed and removed, that is not a very efficient policy.
+Such a policy would place a burden on the (already likely overworked) maintainer team, and the ecosystem would be cluttered with the unwanted packages until they are manually removed.
+
+Allowing a user to delete their package in certain cases is therefore good for the user, the package manager maintainer team, and the package ecosystem at large.
+But what factors need to be considered to avoid a `left-pad` style incident?
+
+## Ecosystem Impact
```
joelverhagen, line 39: Seems like another top- level section is scope of deletion. On NuGet you can delete an entire package name, or just one or more versions.

A policy could vary on whether the latest version is being deleted or whether all the versions are being deleted, etc.

That maybe you are already saying this and I am misunderstanding terminology. It might be worth qualifying at the top of the document that a package refers to all of the versions and a version refers to a single release. Or something like that!

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+Going forward, the user knows that not only can they not expect any new features, but they can't even expect continued compatibility with other projects or security updates.
+
+Similar to version yanking policies, deprecation policies could use slightly different languages including terms like 'end of life' or mention moving to the next major version.
+Regardless of the terms used, with a deprecation policy:
+
+1. The deprecated version/package is still available for installation.
+2. The user is informed of the deprecation and can make an informed decision about continuing to use the package or version.
+3. The build will not break.
+
+## A Comprehensive Policy
```
joelverhagen, line 165: Some other ideas we can consider here: 

Have a public list of deleted packages allowing auditors, tooling, proxies or replicas to react or propagate the deletion. I believe GH has/had a public DMCA list that i think we all should respect 🙂 (it is inspirational).

Recommend having deletion policies qualified on illegal versus malicious versus author request. At least for NuGet some of the considerations go away if the content is illegal. We will delete popular illegal content.

joelverhagen, line 165: Deletion intersects with package immutability as well (which is its own topic of course). But it might be worth recommending that deleted package versions are not replaced and a new release should be made. Or at least a decision point to consider.

For NuGet, we consider a package name + version as immutable and hard deletion of a version opens up the door for package mutability, which can break tooling/caching/replication assumptions.

joelverhagen:
@hayleycd - this is an awesome doc! I left a bunch of feedback. I don't have any big concerns except around the word "yanking". I think the term should be disambiguated at the top of the document or a more ecosystem-neutral term be selected. A terminology section might help to align readers from various backgrounds on a set of shared concepts.

Great work!

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
+# Crafting a Package Deletion Policy
+
+Author: [Hayley Denbraver](https://www.github.com/hayleycd)
+Last updated: April 2025
+
+## Background
+
+In 2016 after a dispute about the naming of an open source package, a maintainer deleted all the packages he had published to npm, including a small package called `left-pad`.
```
joelverhagen, line 8: RE: my point about shared terminology -- it might be good to define the personas/actor names. On NuGet we have:
- site admin
- package owner (sometimes package author)
- package consumer

The term `maintainer` seems to be a little overloaded in this document. Perhaps
- registry maintainer
- package maintainer
- package consumer

But I am not sure. I guess my point goes back to well-defined roles with unambiguous names used throughout the doc.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+For example PyPI maintains [TestPyPI](https://test.pypi.org/) to allow users to try distribution tools and processes without affecting the real PyPI.
+A test instance does not necessarily eliminate the need for a package deletion policy because mistakes can still be made, but it does allow for users to experiment without leaving junk packages on the main instance.
+
+## Deletion Alternatives
+
+There are a number of reasons that a person would want to delete a package.
+Package deletion is a good solution for some of those reasons (publishing the wrong package, making a typo, experimenting with the package manager), but not necessarily for others.
+
+What if a package (or version series) is no longer in active development?
+What if a specific version has a security issue?
```
joelverhagen, line 126:
```suggestion
What if a specific version has a security issue or a severe bug?
```

We have received many deletion requests for a bugged release. We generally recommend unlisting (seems akin to yanking, as you have described below).

docs/package-deletion-policies.md
hayleycd: Yeah, "yanking" is an alternative to deletion. So I would agree, malicious, illegal, or copyrighted content shouldn't be "yanked" but removed entirely.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+Prior to removal, it had over 15 million downloads and was used by thousands of open source projects--including many that are critical to the npm ecosystem.
+The deletion of `left-pad` caused build failures for projects large and small and rendered a large portion of the internet unnavigable. Npm made the decision to republish the package.
+
+[After the `left-pad` incident](https://en.wikipedia.org/wiki/Npm_left-pad_incident), npm updated their package deletion policy to prevent similar failures and other package managers followed suit.
+In the over 9 years since the incident, package managers have continued to adjust their package deletion policies to balance the sometimes competing interests of open source maintainers, open source consumers, and the package ecosystem itself.
+
+This paper outlines the components of modern package deletion policies.
+It is not meant to be a sample deletion policy, but a list of things to consider when crafting such a policy including package deletion alternatives.
+
+## Deletion Policy Considerations
```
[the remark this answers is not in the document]
di: I think this is worth mentioning! (e.g. on PyPI, a project cannot be owner-less, so there is a difference there)

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,218 @@
...
+Similar to version yanking policies, deprecation policies could use slightly different languages including terms like 'end of life' or mention moving to the next major version.
+Regardless of the terms used, with a deprecation policy:
+
+1. The deprecated version/package is still available for installation.
+2. The user is informed of the deprecation and can make an informed decision about continuing to use the package or version.
+3. The build will not break.
+
+## A Comprehensive Policy
+
+A package deletion policy should take into account the needs of open source maintainers, open source consumers, the package registry maintainer team, and the ecosystem at large.
```
di, line 175:
```suggestion
A package deletion policy should take into account the needs of open source maintainers, open source consumers, the package registry administrators, and the ecosystem at large.
```
```suggestion
A package deletion policy should take into account the needs of open source maintainers, open source consumers, the package registry administration team, and the ecosystem at large.
```

docs/package-deletion-policies.md
di: Looks like @hayleycd has mostly disambiguated the term 'maintainer' here. This makes me think the WG should have a general glossary, and align it's various docs around it, but that's probably out of scope for this project.

docs/package-deletion-policies.md
di: Agreed, a section about the various options for "granularity of deletion" would make sense to include, but maybe not in this section.

docs/package-deletion-policies.md
joelverhagen: I was thinking the same thing. a mega synonym table mapping concept across ecosystem would be so nerdy, and so cool

di:
I think this is getting pretty close to merge-able! 

Last call for reviews, and for previous reviewers to resolve open comments -- I'll look to merge this early next week.

docs/package-deletion-policies.md
di: Moving this to https://github.com/ossf/wg-securing-software-repos/issues/52, thanks!

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,218 @@
...
+A package registry administrator team could choose whatever threshold they like, but many package registries choose to not allow for deletion of dependent packages at all.
+
+### Maintainer Status
+
+Many package deletion policies will not allow packages with more than one maintainer to be deleted.
+Allowing a single maintainer to delete the package could contradict the will of the other maintainers and there is not an uncomplicated way to get the consent of the maintainer team.
+Multiple maintainers also implies an increased investment in the project.
+Additionally a package registry could differentiate between a "maintainer" and an "owner" and only allow for the package owner to delete the package.
+
+## What to do About Namespaces?
```
simi, line 96: Term namespace is confusing to me. It is used in various contexts across package repositories. Is there alternative?

simi: added to #52 also

hayleycd: You know, I think that is a good point. I am going to rephrase.

docs/package-deletion-policies.md
```diff
@@ -0,0 +1,210 @@
...
+
+Download counts are not a perfect way to account for the package's impact on the ecosystem.
+They can be arbitrarily inflated and might vary over time.
+Download counts are best used in conjunction with other criteria in evaluating whether a given package should be eligible for deletion.
+
+### Dependency Status
+
+Is the package under consideration a dependency of other packages within the ecosystem?
+How many?
+Is _that_ package a dependency of another package (making the package to be deleted a transitive dependency)?
```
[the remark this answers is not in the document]
hayleycd: That makes a lot of sense, but I guess I would be curious about the download numbers of those packages, etc. To me this feels like a circumstance where they might need to reach out to the admins directly. I am leaving it out for now, but if it is really important to the team, I can put it in.

docs/package-deletion-policies.md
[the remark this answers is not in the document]
hayleycd: I made some changes to address this.

docs/package-deletion-policies.md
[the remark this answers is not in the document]
joelverhagen: For NuGet the download count of dependency packages is incremented along with the parent packages. This means if there is a core package in an isolated subgraph like "MyObsurePackageGraph.Core", it might have a relatively high download count but in fact just be pumped by the size of the graph moreso than the popularity of the package. 

I think the wording is fine as-is since the depth of package graphs is probably an ecosystem specific thing, and whether subgraphs are often published by a single author is also ecosystem specific. 

Feel free to resolve! And thanks for hearing me out :)

di:
Thank you @hayleycd!

PR #57, opened 2025-05-17
fix: replace job_workflow_ref with workflow_ref

woodruffw:
TL;DR: `job_workflow_ref` and `workflow_ref` are
*often* the same thing and have the same value,
but sometimes diverge in ways that make later
support for GitHub's reusable workflows difficult.

This guide should recommend `workflow_ref` instead of `job_workflow_ref` for the "baseline" of Trusted
Publishing, since it's always correct as the
"initiating" workflow identity.

See https://github.com/pypi/warehouse/issues/11096 and https://github.com/rust-lang/crates.io/pull/11131#discussion_r2083661244 for more context.

CC @sethmlarson


sethmlarson (approved):
LGTM, thanks Will!

PR #96, opened 2025-11-13
Fix image links that were broken from the style guide

rmhartog:
Some images were linked from articles, but were moved into the style guide images directory, even though they are unrelated. This change moves them back to their original directory.


di (approved):
Thanks!

PR #97, opened 2025-12-18
Update list of registries that support trusted publishing

36degrees:
Add three more registries that are known to support trusted publishing:

- npm
- crates (Rust)
- NuGet (.net)

This is not an attempt to create an exhaustive list, just to add registries that I am personally aware of.

As the list of registries has grown, split it out from a paragraph of inline links into a bulleted list.

Move the list into a new 'Support' section to differentiate it from the list of other references that previously followed the paragraph.


di (approved):

di:
Thanks!

